PR #2, opened 2024-11-12
Fix pytest action triggers

johannesbulin:
Fixed the triggers in the `pytest.yaml` action such that it may run in the following circumstances:
* push to `main`.
* Changes to a pull request.
* Manually triggered.


mlange05 (approved):
Yes, looks good to me. :shipit:

PR #5, opened 2024-12-11
Python 3.12+ compatibility and add code coverage to pytest

reuterbal:
I replaced the use of pkg_resources with importlib for compatibility with Python 3.12+. This is backward compatible until 3.8, which we require as a minimum already.

I've also adapted the pytest plan to include code coverage and upload this to codecov. Unfortunately, there was no easy way to hook this into the reusable workflow, but given the very few extra lines needed for equivalent functionality, it seemed easiest to do it this way.

### DIFF
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -13,29 +13,59 @@ on:
   workflow_dispatch:
 
 jobs:
-    pylint:
-        runs-on: ubuntu-latest
+  pylint:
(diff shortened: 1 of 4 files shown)


mlange05 (approved):
Looks good to me. Many thanks for the TLC and additional codecov covergage. GTG form me. :shipit:

johannesbulin (approved):
I agree, changes look nice. So thanks for adding this!

Issue #6, opened 2024-12-16
ifsbench.files for data pipeline


johannesbulin:
### Is your feature request related to a problem? Please describe.

The data handling in `ifsbench` is currently limited to extracting an archive and moving some files around. In the future we may need a more extensive _pipeline_ like approach to support things like
- different namelist/file modifications, depending on the application that is run.
- options at runtime that modify some namelist entries to enable/disable additional debugging or GPU capabilities.

### Describe the solution you'd like

Implement a `DataHandler` class that performs a given operation in a working directory. It should be possible to execute several `DataHandler` sequentially (in a pipeline manner), for example
1. Extract an archive with raw input data
2. Move some files around to adapt to a given benchmark.
3. Update namelist entries.

To do this, implement the `DataHandler` base class and specialisations for

- archive extraction
- renaming/copying
- namelist modifications

### Organisation

ECMWF

PR #7, opened 2024-12-16
#6: Implemented DataHandler class and extract/remove/namelist

johannesbulin:
(See issue #6)

Added support for data pipelines to `ifsbench`. The main feature is a `DataHandler` base class that can perform a given action on a working directory. Currently I've implemented:

- `RenameHandler` - for moving/renaming/copying files, using regular expressions.
- `NamelistHandler` - for modifying namelists.
- `ExtractHandler` - for extracting archives.

Talking points / request for feedback:

- At the moment, nothing in the data pipeline uses the `InputFile` concept that we've used before (with file hashes and other things). If you think that this should be included, I'd be happy about some suggestions.
- I am not a big fan of having enums as embedded classes (`RenameHandler.RenameMode` for example). It's a bit clunky to write but I haven't come up with a better solution.

### DIFF
--- a/ifsbench/data/extracthandler.py
+++ b/ifsbench/data/extracthandler.py
@@ -1,0 +1,53 @@
+# (C) Copyright 2020- ECMWF.
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+
+import pathlib
+import shutil
+
+from .datahandler import DataHandler
+from ..logging import debug
+
+__all__ = ['ExtractHandler']
+
+
+class ExtractHandler(DataHandler):
+    """
+    DataHandler that extracts a given archive to a specific directory.
+    """
+
+    def __init__(self, archive_path, target_dir=None):
+        """
+        Initialise the handler.
+
+        Parameters
+        ----------
+        archive_path: str or `pathlib.Path`
+            The path to the archive that will be extracted. If a relative path
+            is given, this will be relative to the `wdir` argument in `execute`.
+
+        target_dir: str, `pathlib.Path` or `None`
+            The directory to where the archive will be unpacked. If a relative path
+            is given, this will be relative to the `wdir` argument in `execute`.
+            If None is given, this will re extracted to `wdir`.
+        """
+
+        self._archive_path = pathlib.Path(archive_path)
+        if target_dir is None:
+            self._target_dir = None
+        else:
+            self._target_dir = pathlib.Path(target_dir)
+
+    def execute(self, wdir, **kwargs):
+        target_dir = wdir
+        if self._target_dir is not None:
+            if self._target_dir.is_absolute():
+                target_dir = self._target_dir
+            else:
+                target_dir = wdir/self._target_dir
+
+        debug(f"Unpack archive {self._archive_path} to {target_dir}.")
+        shutil.unpack_archive(self._archive_path, target_dir)
--- a/ifsbench/data/extracthandler.py
+++ b/ifsbench/data/extracthandler.py
@@ -8,8 +8,8 @@
 import pathlib
 import shutil
 
-from .datahandler import DataHandler
-from ..logging import debug
+from ifsbench.data.datahandler import DataHandler
+from ifsbench.logging import debug
 
 __all__ = ['ExtractHandler']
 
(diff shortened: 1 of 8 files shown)


FussyDuck:
[CLA assistant check](https://bol-claassistant.ecmwf.int:443/ecmwf-ifs/ifsbench?pullRequest=7) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://bol-claassistant.ecmwf.int:443/ecmwf-ifs/ifsbench?pullRequest=7) before we can accept your contribution.<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://bol-claassistant.ecmwf.int:443/check/ecmwf-ifs/ifsbench?pullRequest=7) it.</sub>

ifsbench/data/namelisthandler.py
```diff
@@ -0,0 +1,167 @@
...
+            if self._mode == self.NamelistOperation.DELETE:
+                return
+
+            namelist[self._keys[0]] = {}
+
+        namelist = namelist[self._keys[0]]
+        key = self._keys[-1]
+
+        if self._mode == self.NamelistOperation.SET:
+            debug(f"Set namelist entry {str(self._keys)} = {str(self._value)}.")
```
reuterbal, line 83: [Purely FYI, no action required] The following should be a string conversion shorthand for f-strings

```suggestion
            debug(f"Set namelist entry {self._keys!s} = {self._value!s}.")
```

ifsbench/data/namelisthandler.py
```diff
@@ -0,0 +1,167 @@
...
+                debug(f"Delete namelist entry {str(self._keys)}.")
+                del namelist[key]
+
+class NamelistHandler(DataHandler):
+    """
+    DataHandler specialisation that can modify Fortran namelists.
+    """
+
+    def __init__(self, input_path, output_path, overrides):
+        """
```
reuterbal, line 119: Move the Parameters section to the class docstring please

ifsbench/data/namelisthandler.py
```diff
@@ -0,0 +1,167 @@
...
+            The NamelistOverrides that will be applied.
+        """
+
+        self._input_path = pathlib.Path(input_path)
+        self._output_path = pathlib.Path(output_path)
+
+        self._overrides = list(overrides)
+        for override in self._overrides:
+            if not isinstance(override, NamelistOverride):
+                raise ValueError("Namelist overrides must be NamelistOverride objects!")
```
reuterbal, line 143: You could use a list comprehension here:
```suggestion
        if not all(isinstance(override, NamelistOverride) for override in self._overrides):
            raise ValueError("Namelist overrides must be NamelistOverride objects!")
```

ifsbench/data/renamehandler.py
```diff
@@ -0,0 +1,114 @@
...
+            The pattern that will be replaced. Corresponds to `pattern` in
+            `re.sub`.
+
+        repl: str
+            The replacement pattern. Corresponds to `repl` in `re.sub`.
+
+        mode:   `RenameHandler.RenameMode`
+            Specifies how the renaming is done (copy, move, symlink).
+
+        mode:   `RenameHandler.RenameMode`
```
reuterbal, line 58:
```suggestion
        mode: RenameHandler.RenameMode
```

tests/data/test_renamehandler.py
```diff
@@ -0,0 +1,97 @@
...
+        ['data.txt', 'data1.txt', 'data2.txt'],
+    ),
+    (r'data[12]/', r'data/', RenameHandler.RenameMode.MOVE,
+        ['data/data.txt','data1/data.txt', 'data1/data2.txt'],
+        ['data/data.txt', 'data/data2.txt'],
+    ),
+    (r'replacement$', r'dummypath', RenameHandler.RenameMode.COPY,
+        ['dummypath/somedata.tar.gz','replacement'],
+        ['dummypath', 'replacement'],
+    ),
```
reuterbal, line 48: `repl` should not be an r-string, right?

Also, if you decide to allow also compiled patterns I would add or convert 1-2 of the parameterisations to use `re.compile(r'...')` as `pattern`

reuterbal (changes requested):
Thanks, this looks great already!

I left a lot of nitpicky comments but mostly around style and docstrings. Implementation looks mostly good to me and testing is very diligently done!

ifsbench/data/extracthandler.py
```diff
@@ -0,0 +1,53 @@
...
+            If None is given, this will re extracted to `wdir`.
+        """
+
+        self._archive_path = pathlib.Path(archive_path)
+        if target_dir is None:
+            self._target_dir = None
+        else:
+            self._target_dir = pathlib.Path(target_dir)
+
+    def execute(self, wdir, **kwargs):
```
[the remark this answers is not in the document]
johannesbulin: Does the `execute` function needs its own docstring? I thought that it would just inherit the docstring from the parent class `DataHandler`?!

ifsbench/data/renamehandler.py
```diff
@@ -0,0 +1,114 @@
...
+        Parameters
+        ----------
+        pattern: str
+            The pattern that will be replaced. Corresponds to `pattern` in
+            `re.sub`.
+
+        repl: str
+            The replacement pattern. Corresponds to `repl` in `re.sub`.
+
+        mode:   `RenameHandler.RenameMode`
```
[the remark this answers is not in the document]
johannesbulin: Went for :class:`RenameMode`. Hope this is okay as well.

tests/data/test_renamehandler.py
johannesbulin: Tried to change this but now it complains about repl strings that contain `\g<something>`. So I guess that at least some of them should should be r-strings.

ifsbench/data/extracthandler.py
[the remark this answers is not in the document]
reuterbal: Not sure if it works like this but fine with keeping it like this for now.

reuterbal (approved):
Many thanks for making the suggested changes! Looks great to me now.

PR #10, opened 2025-01-27
Config dict

uhager:


FussyDuck:
[CLA assistant check](https://bol-claassistant.ecmwf.int:443/ecmwf-ifs/ifsbench?pullRequest=10) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://bol-claassistant.ecmwf.int:443/ecmwf-ifs/ifsbench?pullRequest=10) before we can accept your contribution.<br/><hr/>**Ulrike Hager** seems not to be a GitHub user. You need a GitHub account to be able to sign the CLA. If you have already a GitHub account, please [add the email address used for this commit to your account](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user).<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://bol-claassistant.ecmwf.int:443/check/ecmwf-ifs/ifsbench?pullRequest=10) it.</sub>

ifsbench/config_mixin.py
```diff
@@ -0,0 +1,120 @@
...
+from typing import Any, get_args, get_origin, get_type_hints, Optional, TypeVar, Union
+
+__all__ = ['ConfigMixin', 'CONF']
+
+
+CONF = Union[int, float, str, bool, dict, list, None]
+
+
+def _config_from_locals(config: dict[str, Any]) -> None:
+    print(f'from locals: config={config}, type={type(config)}')
```
johannesbulin, line 11: Just a comment for later (I now that this is a draft). Most output in `ifsbench` is done via the `debug/info/warning/....` functions in `ifsbench.logging`.

ifsbench/config_mixin.py
```diff
@@ -0,0 +1,120 @@
...
+            raise ValueError('Config already set.')
+        self._config = config
+
+    def get_config(self) -> dict[str, CONF]:
+        return self._config
+
+    def update_config(self, field: str, value: CONF) -> None:
+        if field not in self._config:
+            raise ValueError(f'{field} not part of config {self._config}, not setting')
+        if type(value) != type(self._config[field]):
```
johannesbulin, line 62: Type comparisons might be dangerous here, if we work with subclasses (for example if a class stores multiple `DataHandler` objects in a list).

ifsbench/config_mixin.py
```diff
@@ -0,0 +1,120 @@
...
+
+    The required format can be either created based on the constructor, or explicitly set by
+    implementing the `config_format` method.
+
+    Parameters
+    ----
+    config: dictionary containing parameter names and their values
+    """
+
+    _config = None
```
johannesbulin, line 34: Should probably be given a less generic name like `_mixin_config`. Otherwise this might clash with classes that have their own `_config` attribute (which is probably not that rare).

johannesbulin (commented):
I'm quite happy with the idea of using a mixin. The code is (as usual) well written, simple and clean.

I have a few concerns, mainly regarding the use of type hints:
- As you can see in the checks, the currently used type hints don't work for all Python versions (major changes were added in Python 3.9, some smaller later). I don't know whether we can make it work for all versions.
- In my opinion, type hints are hints - and are usually not enforced.

This is probably not a blocking issue. Unless I am mistaken, we could simply check the types at (de-)serialisation time.

One open question: How would you deserialise an object (i.e. from YAML to a Python object). I couldn't find any references to the Python objects class/module yet and it would also need some kind of `from_dict` class function to recover an object.

mlange05:
@uhager First of all, many thanks for looking into this, and even though it's still a draft, I think this is a very promising start for a complex design issue. I've not looked through this in detail, but the initial discussion is already super interesting and valuable. So, thanks for your initiative! 🙏 

@uhager and @johannesbulin I agree on both accounts that the Mixin approach is great, and that using type-info for the necessary introspection are very neatly done. 

On @johannesbulin point about type-safety and automated checking, I would like to point out that there is also the option to use Pydantic: https://docs.pydantic.dev/latest/
This is a library that performs strong type checking against Python type hints, and allows various levels of control over strictness and automatic type-coaxing. We use it in one of the lower layers of Loki, and we have Python version support for Python 3.8-3.12 (3.13 is missing for other reasons). What I also find convenient about Pydantic is the use of a `BaseModel` abstraction that allows you to specify the types as class properties, rather than type-hints to the constructor. This comes with neat features like "validators" that can be used to customise our own "auto-conversion" of types, if we need to.

And lastly, I'm open to either solution, but as type-safety and introspection were raise, I thought I'd throw out some pointers.

ifsbench/config_mixin.py
uhager: Done.

ifsbench/config_mixin.py
uhager: Changed to isinstance

ifsbench/config_mixin.py
uhager: This was just a debug output I forgot. Removed now

uhager:
> One open question: How would you deserialise an object (i.e. from YAML to a Python object). I couldn't find any references to the Python objects class/module yet and it would also need some kind of `from_dict` class function to recover an object.

I think that should be a separate function / class with implementations for yaml, json, or whatever we want to support. That function will parse the yaml, create the config dicts, and call the build functions in the classes that need to be created

johannesbulin:
> > One open question: How would you deserialise an object (i.e. from YAML to a Python object). I couldn't find any references to the Python objects class/module yet and it would also need some kind of `from_dict` class function to recover an object.
> 
> I think that should be a separate function / class with implementations for yaml, json, or whatever we want to support. That function will parse the yaml, create the config dicts, and call the build functions in the classes that need to be created

Fine with the separate function to start the whole (de-)serialising process. Still, if you have a class like this
```
class SomeBenchmark(ConfigMixin):
   def __init__(self, data_handler):
       # data_handler is list of DataHandler objects.
       self._data_handler = list(data_handler)
```
you would have `ConfigMixin` objects (in data_handler) encapsulated in another `ConfigMixin` object (`SomeBenchmark` instance). How would you serialise this? If I understand this correctly, `config_format` in `SomeBenchmark` should also call `config_format` in the `data_handler` objects. 
And at this point you would also have to store information on classes/modules.

uhager:
> On @johannesbulin point about type-safety and automated checking, I would like to point out that there is also the option to use Pydantic: 

Looked into that, looks promising. I have never used pydantic, so to see what's possible I made a prototype implementation: https://github.com/ecmwf-ifs/ifsbench/pull/13

That approach is a lot cleaner than this one, but may end up less flexible. So far it's looking good though.

Issue #11, opened 2025-01-27
ifsbench rework launching mechanism


johannesbulin:
### Is your feature request related to a problem? Please describe.

At the moment, essentially he classes

- `Job`
- `Arch`
- `Launcher`

are responsible for launching executables. As these are pure classes, they can't easily be customised but most be subclassed. They also don't interact nicely with the new datapipeline approach.

### Describe the solution you'd like

- `Arch` and `Launcher` classes are overhauled. They become abstract base classes and simple subclasses are implemented to support the most basic needs. 
- System specific solutions are removed (no LUMI arch files for example).
- Similar to the data pipeline approach, an environment pipeline should be created.

### Organisation

ECMWF

PR #12, opened 2025-01-28
Reimplemented launching procedure (#11).

johannesbulin:
Updated launching mechanism (see #11)

- Added environment handler and environment pipeline classes.
- Added `Application` base class.
- `Arch` classes now have methods instead of class methods. Will now update `Job` objects and may create additional `EnvHandler` objects.
- Reworked `Launcher` classes.
- Removed `Benchmark` classes and tests as these are broken due to the extensive changes and will be reimplemented later.


ifsbench/application.py
```diff
@@ -0,0 +1,149 @@
...
+            The working directory from where the executable will be launched.
+        job: Job
+            The parallel job setup.
+
+        Returns
+        -------
+        list[DataHandler]:
+            List of DataHandler objects.
+        """
+        return NotImplemented
```
reuterbal, line 37: I don't think the explicit return statements are required. A derived class that doesn't implement these methods can't be instantiated, and therefore this line can never be reached, [see](https://docs.python.org/3/library/abc.html#abc.abstractmethod):
> Using this decorator requires that the class’s metaclass is [ABCMeta](https://docs.python.org/3/library/abc.html#abc.ABCMeta) or is derived from it. A class that has a metaclass derived from ABCMeta cannot be instantiated unless all of its abstract methods and properties are overridden.

ifsbench/env.py
```diff
@@ -0,0 +1,197 @@
...
+            Handler(s) that are added.
+        """
+
+class DefaultEnvPipeline(EnvPipeline):
+    """
+    Default environment pipeline.
+
+    Parameters
+    ----------
+    overrides: list[EnvHandler]
```
reuterbal, line 158: This looks like an outdated docstring
```suggestion
    handlers: list[EnvHandler]
```

ifsbench/env.py
```diff
@@ -0,0 +1,197 @@
...
+        env: dict
+            An environment dictionary. Will be updated in place.
+        """
+
+        if self._mode == EnvOperation.SET:
+            debug(f"Set environment entry {self._key} = {self._value}.")
+            env[self._key] = self._value
+        elif self._mode == EnvOperation.APPEND:
+            if env.get(self._key, None) is not None:
+                env[self._key] += ":" + self._value
```
reuterbal, line 99: For portability, you might want to consider using [`os.pathsep`](https://docs.python.org/3/library/os.html#os.pathsep) here (and in the other relevant places)
```suggestion
                env[self._key] += os.pathsep + self._value
```

ifsbench/job.py
```diff
@@ -106,206 +90,134 @@ class CpuDistribution(Enum):
...
-    def get_tasks(self):
-        """
-        The total number of MPI tasks
+    tasks : int = None
+    nodes : int = None
+    tasks_per_node : int = None
+    tasks_per_socket : int = None
+    cpus_per_task : int = None
+    threads_per_core : int = None
+    gpus_per_task : int = None
```
reuterbal, line 133: We should consider whether we want to only use `gpus_per_task` here, or have an option to also use `gpus_per_node`. Reason being that with MPS on Nvidia you can map multiple MPI ranks to the same GPU. This allows you to run more MPI tasks per node than you have GPUs, which can be beneficial if your host code doesn't scale well with thread parallelism. But this likely also invalidates the concept of `gpus_per_task` as this would have to become something like 1/2, 1/3, 1/4?

ifsbench/launcher/__init__.py
```diff
@@ -0,0 +1,11 @@
...
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+
+from .launcher import * # noqa
+from .bashlauncher import * # noqa
+from .mpirunlauncher import * # noqa
+from .srunlauncher import * # noqa
```
reuterbal, line 11: We should switch to absolute imports here

ifsbench/launcher/bashlauncher.py
reuterbal, line 1: With Python's de-facto standard snake_case naming scheme, this should be called `bash_launcher.py` (similarly for the other files)

ifsbench/launcher/bashlauncher.py
```diff
@@ -0,0 +1,75 @@
...
+# nor does it submit to any jurisdiction.
+
+import datetime
+from pathlib import Path
+
+from ifsbench.launcher.launcher import Launcher, LaunchData
+
+class BashLauncher(Launcher):
+    """
+    UNTESTED!
```
reuterbal, line 15: Can we defer integrating this until we have tests for this?

ifsbench/launcher/bashlauncher.py
```diff
@@ -0,0 +1,75 @@
...
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+
+import datetime
+from pathlib import Path
+
+from ifsbench.launcher.launcher import Launcher, LaunchData
+
+class BashLauncher(Launcher):
```
reuterbal, line 13: Conceptually, I'd argue it would be better to make this a `ScriptLauncher` or `ShellScriptLauncher` and let the fact that this uses Bash be an implementation detail.

ifsbench/launcher/bashlauncher.py
```diff
@@ -0,0 +1,75 @@
...
+        timestamp = current_datetime.strftime('%Y-%m-%d.%H:%M:%S.%f')
+
+        # Guess the executable name from the command.
+        exec_name = cmd[0].split('/')[-1]
+        path = script_dir/f"run-{exec_name}-{timestamp}.sh"
+
+        run_env = child_data.env
+
+        with path.open('w') as f:
+            f.write('#! /bin/bash')
```
reuterbal, line 61: The white space looks odd?!
```suggestion
            f.write('#!/bin/bash')
```

ifsbench/launcher/launcher.py
```diff
@@ -0,0 +1,91 @@
...
+        debug("Environment variables:")
+        for key, value in self.env.items():
+            debug(f"\t{key}={value}")
+
+
+        execute(
+            command = self.cmd,
+            cwd = self.run_dir,
+            env = self.env
+        )
```
reuterbal, line 57: This appears to be untested

tests/launcher/test_srunlauncher.py
```diff
@@ -0,0 +1,77 @@
...
+import pytest
+
+from ifsbench import (Job, SrunLauncher, EnvHandler, EnvOperation,
+    DefaultEnvPipeline, CpuBinding, CpuDistribution)
+
+_test_env = DefaultEnvPipeline([
+        EnvHandler(EnvOperation.SET, 'SOME_VALUE', '5'),
+        EnvHandler(EnvOperation.SET, 'OTHER_VALUE', '6'),
+        EnvHandler(EnvOperation.DELETE, 'SOME_VALUE'),
+])
```
reuterbal, line 21: This would be better as a fixture

tests/launcher/test_mpirunlauncher.py
```diff
@@ -0,0 +1,78 @@
...
+
+from ifsbench import (Job, MpirunLauncher, EnvHandler, EnvOperation,
+    DefaultEnvPipeline, CpuBinding, CpuDistribution)
+
+
+_test_env = DefaultEnvPipeline([
+        EnvHandler(EnvOperation.SET, 'SOME_VALUE', '5'),
+        EnvHandler(EnvOperation.SET, 'OTHER_VALUE', '6'),
+        EnvHandler(EnvOperation.DELETE, 'SOME_VALUE'),
+])
```
reuterbal, line 22: This would be better as a fixture

ifsbench/job.py
```diff
@@ -106,206 +90,134 @@ class CpuDistribution(Enum):
...
-        The total number of nodes
 
-        If this has not been specified explicitly, it is estimated as
-        ``ceil(threads / available_threads_per_node)`` with the number of
-        available threads dependent on the use of SMT.
+        return replace(self)
+
+
+
+    def calculate_missing(self, cpu_configuration):
```
reuterbal, line 149: If we used pydantic, this could maybe be implemented via field_validators or similar

ifsbench/job.py
```diff
@@ -106,206 +90,134 @@ class CpuDistribution(Enum):
     DISTRIBUTE_USER = auto()
     """Indicate that a different user-specified strategy should be used"""
 
+@dataclass
```
reuterbal, line 93: I like the idea of dataclasses here and elsewhere. Should we maybe combine this directly with pydantic?

ifsbench/arch.py
```diff
@@ -9,467 +9,144 @@
...
-            The number of GPUs that are used per MPI task
-        kwargs :
-            Other arguments that may be used in the architecture implementation
-            or may be passed on to :any:`execute`
+        Returns
+        -------
+        Launcher
+            The default launcher used in this architecture.
         """
+        return NotImplemented
```
reuterbal, line 62: Return values for abstract methods can likely be skipped again

reuterbal (changes requested):
Many thanks, conceptually this looks very promising!

I've left a few remarks, most are minor but a few conceptual questions. In particular, I'd suggest to evaluate to what degree we could combine the use of dataclasses (which I love, BTW) with pydantic - the introduction of the latter into Loki has reduced the error rate dramatically and therefore it's likely a very valuable addition here, too.

Finally, due to #14 we have conflicts that need to be resolved.

VERSION
```diff
@@ -1 +1 @@
-0.1.2
+0.2.0
```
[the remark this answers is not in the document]
johannesbulin: Agree. And done...

ifsbench/__init__.py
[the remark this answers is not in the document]
johannesbulin: Rebased and resolved the conflicts.

ifsbench/launcher/bashlauncher.py
johannesbulin: Yes, I've taken it out for now.

ifsbench/launcher/launcher.py
johannesbulin: Added simple test for this.

ifsbench/job.py
johannesbulin: I'd be happy to support `gpus_per_node` in the future. In my opinion, this should be outsourced to a separate issue/PR as this would require multiple changes. 

From my point of view, `Job` is essentially just a dataclass/dict-like-object that holds all user-specified parallel options with (almost) no logic. So adding `gpus_per_node` in addition to `gpus_per_task` should be no big deal. We would just have to decide which part of the code would be responsible to handle clashes in the job attributes. We could do this in the `Launcher` classes or just decide that all options are forwarded to the actual launcher executables `srun/mpirun` and if they are incompatible it's not `ifsbench`'s responsibility.

As I said, this needs a bigger discussion...

ifsbench/job.py
johannesbulin: I'd be happy to use pydantic in more places here - it's already used in the `ConfigMixin` class. As this development took place at the same time as these launcher changes, I would prefer to merge these changes as they are and do some additional pydantic improvements afterwards. Otherwise I'll end in some kind of rebase/testing hell.

tests/launcher/test_srunlauncher.py
johannesbulin: Just tried to fix it - `pytest.fixture` doesn't work in a `pytest.mark.parametrize` setting. I couldn't find a good solution for this immediately - do you have any idea?

ifsbench/job.py
reuterbal: Yes absolutely - sorry, should have made clear that this is out of scope here.

tests/launcher/test_srunlauncher.py
johannesbulin: Thanks! Works nicely!

ifsbench/application.py
```diff
@@ -0,0 +1,149 @@
...
+
+__all__ = ['Application', 'DefaultApplication']
+
+class Application(ABC):
+    """
+    Base class for applications that can be launched.
+    """
+
+    @abstractmethod
+    def get_data_handlers(self, run_dir, job):
```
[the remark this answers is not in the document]
johannesbulin: I think that this would be useful. If possible, I'd like to do this in a separate pull request which would also update the files that haven't been touched here.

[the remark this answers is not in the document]
uhager: I can also do that if you'd like. Might be useful for me to better understand the code.

ifsbench/arch.py
```diff
@@ -9,467 +9,140 @@
...
+        flags.
+    default_launcher: Launcher
+        The default launcher that is used on this system.
+    default_launcher_flags : list[str]
+        Additional launcher flags that should be added to launcher invocations.
     """
+    job = None
+    env_handlers : List[EnvHandler] = field(default_factory=list)
+    default_launcher : Launcher = None
+    default_launcher_flags : List[str] = field(default_factory=list)
```
reuterbal, line 39: Docstrings won't render right, here.

```suggestion
    """
    #: The updated job after the architecture processing.
    job = None
    
    #: Additional EnvHandler objects that set architecture-specific environment flags
    env_handlers : List[EnvHandler] = field(default_factory=list)
    
    #: The default launcher that is used on this system.
    default_launcher : Launcher = None
    
    #: Additional launcher flags that should be added to launcher invocations.
    default_launcher_flags : List[str] = field(default_factory=list)
```

ifsbench/env.py
```diff
@@ -0,0 +1,197 @@
...
+__all__ = ['EnvHandler', 'EnvOperation', 'EnvPipeline']
+
+
+class EnvOperation(Enum):
+    """
+    Enum of environment operations.
+
+    Specifies operations on environment variables.
+
+    Attributes
```
reuterbal, line 24: Same issue - see correct Docstring for class attributes in previous comment or #18

ifsbench/job.py
```diff
@@ -36,39 +30,27 @@ class CpuConfiguration(ABC):
...
-        from the above properties.
-    threads_per_node : int
-        The number of logical cores per node (threads). This value is automatically derived
-        from the above properties.
     gpus_per_node : int
         The number of available GPUs per node.
     """
 
-    sockets_per_node: int
+    sockets_per_node : int = 1
```
reuterbal, line 37: Same ;-)

ifsbench/job.py
reuterbal: Can you maybe create an issue to track this effort then?

ifsbench/job.py
```diff
@@ -107,206 +89,134 @@ class CpuDistribution(Enum):
...
-    :any:`AttributeError` otherwise.
-
-    Multi-threading can be specified by providing a value larger than 1 (the default)
-    for :data:`cpus_per_task`. Symmetric multi-threading (hyperthreading) can be
-    enabled with a value greater than 1 in :data:`threads_per_core`.
-
-    The desired pinning strategy can be specified with :data:`bind`.
+    Description of a parallel job setup.
 
     Parameters
```
reuterbal, line 97: These are no longer parameters to the constructor, so should be converted to attribute docstrings

reuterbal (changes requested):
Thanks, looks mostly good to me now. I'd appreciate a quick round of docstring fixes, in anticipation of #18.

Also, we have a bunch of unused utilities in ifsbench/util.py now (see codecov report), we should remove them as they are now untested.

ifsbench/job.py
johannesbulin: https://github.com/ecmwf-ifs/ifsbench/issues/19

reuterbal (approved):
Many thanks, all good from my side now!

PR #13, opened 2025-01-29
Config implementation using pydantic

uhager:
I still decided to create a ConfigMixin class so that we can keep the API is we want to move away from pydantic at some point.

### DIFF
--- a/ifsbench/config_mixin.py
+++ b/ifsbench/config_mixin.py
@@ -1,0 +1,17 @@
+from typing import Dict, List, Union
+
+from pydantic import BaseModel
+
+__all__ = ['ConfigMixin']
+
+
+class ConfigMixin(BaseModel):
+
+    @classmethod
+    def from_config(
+        cls, config: Dict[str, Union[str, float, int, bool, List, None]]
+    ) -> 'ConfigMixin':
+        return cls(**config)
+
+    def dump_config(self) -> Dict[str, Union[str, float, int, bool, List]]:
+        return self.model_dump(exclude_none=True)
--- a/ifsbench/config_mixin.py
+++ b/ifsbench/config_mixin.py
@@ -1,3 +1,10 @@
+# (C) Copyright 2020- ECMWF.
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+
 from pathlib import Path
 from typing import Dict, List, Union
 
--- a/ifsbench/config_mixin.py
+++ b/ifsbench/config_mixin.py
@@ -1,3 +1,4 @@
+from pathlib import Path
 from typing import Dict, List, Union
 
 from pydantic import BaseModel, model_validator
@@ -5,6 +6,10 @@ from typing_extensions import Self
 
 __all__ = ['ConfigMixin', 'CLASSNAME', 'RESERVED_NAMES']
 
+# Reserved strings:
+# 'classname' is used in the configuration to indicate which class has to be
+# constructed with that configuration and cannot be used for member variables
+# in implementing classes.
 CLASSNAME = 'classname'
 RESERVED_NAMES = [
     CLASSNAME,
@@ -12,23 +17,47 @@ RESERVED_NAMES = [
 
 
 class ConfigMixin(BaseModel):
+    """
+    Base class for handling configurations in a format that can be used for storage.
+
+    Uses pydantic for type checking and managing the config dictionary.
+    """
 
     @classmethod
     def from_config(
         cls, config: Dict[str, Union[str, float, int, bool, List, None]]
     ) -> 'ConfigMixin':
+        """Create instance based on config.
+
+        Args:
+            config: names and values for member variables.
+
+        Returns:
+            class instance
+        """
         return cls(**config)
 
     def dump_config(
         self, with_class: bool = False
-    ) -> Dict[str, Union[str, float, int, bool, List]]:
+    ) -> Dict[str, Union[str, float, int, bool, Path, List]]:
+        """Get configuration for output.
+
+        Args:
+            with_class: Add CLASSNAME key with class name to configuration.
+
+        Returns:
+            Configuration that can be used to create instance.
+        """
         config = self.model_dump(exclude_none=True)
+        for k, v in config.items():
+            if isinstance(v, Path):
+                config[k] = str(v)
         if with_class:
             config['classname'] = type(self).__name__
         return config
 
     @model_validator(mode='before')
-    def validate_variable_names_not_reserved(self) -> Self:
+    def _validate_variable_names_not_reserved(self) -> Self:
         if any(var in self.keys() for var in RESERVED_NAMES):
             raise ValueError(
                 f'Invalid ConfigMixin class: contains reserved member name(s). Reserved: {RESERVED_NAMES}'
--- a/ifsbench/config_mixin.py
+++ b/ifsbench/config_mixin.py
@@ -1,11 +1,15 @@
 from typing import Dict, List, Union
 
-from pydantic import BaseModel
+from pydantic import BaseModel, model_validator
+from typing_extensions import Self
 
-__all__ = ['ConfigMixin', 'CLASSNAME', 'RESERVED_STRINGS']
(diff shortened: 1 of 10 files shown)


FussyDuck:
[CLA assistant check](https://bol-claassistant.ecmwf.int:443/ecmwf-ifs/ifsbench?pullRequest=13) <br/>Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://bol-claassistant.ecmwf.int:443/ecmwf-ifs/ifsbench?pullRequest=13) before we can accept your contribution.<br/><hr/>**Ulrike Hager** seems not to be a GitHub user. You need a GitHub account to be able to sign the CLA. If you have already a GitHub account, please [add the email address used for this commit to your account](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user).<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://bol-claassistant.ecmwf.int:443/check/ecmwf-ifs/ifsbench?pullRequest=13) it.</sub>

ifsbench/data/namelisthandler.py
```diff
@@ -42,22 +49,17 @@ class NamelistOverride:
         The value that is set (SET operation) or appended (APPEND).
     """
 
+    namelist: str
+    entry: str
```
johannesbulin, line 53: I would be very happy if we could pass both `str` and `Path` here. As far as I know, `pydantic` can handle autoconversion from `Path` to `str`, so I don't even know if what you wrote already supports this.

uhager: pydantic doesn't automatically handle that case, it would require type checks and /or separate fields. 
What is the use case for this? We can't really read a path object directly from yaml, so this would be running from already created objects?
If this is really needed, I can add something.

johannesbulin: Sorry, this should have gone to `NamelistHandler` (where it makes sense) instead of `NamelistOverride` (where this would be completely stupid, I agree...).

ifsbench/data/namelisthandler.py
```diff
@@ -129,23 +138,28 @@ class NamelistHandler(DataHandler):
         The NamelistOverrides that will be applied.
     """
 
-    def __init__(self, input_path, output_path, overrides):
+    input_path: str
```
johannesbulin, line 141: I would be very happy if we could pass both str and Path here. As far as I know, pydantic can handle autoconversion from Path to str, so I don't even know if what you wrote already supports this.

ifsbench/data/namelisthandler.py
uhager: oh, I didn't even notice, I assumed you meant NamelistHandler for the input_path / output_path. 
What's the use-case for this? I think I may still be a bit confused as to what we're trying to do. If these objects are only created from config, supporting Path doesn't really make sense. If we want to create them from other already instantiated objects, the yaml_reader doesn't really make sense. If there is a plan to create them from other objects like Path and then output the yaml config with strings?

johannesbulin: While we want most of the stuff to be (de-)serialisable, all the classes should also work nicely from the Python side. As many parts of `ifsbench` use `pathlib.Path` instead of `str`, it would be great if the different parts were compatible. 
To shorten this discussion: I am happy with keeping this as a `str` for the moment. If the need arises, we can update this in a later pull request (probably we could already pass a `pathlib.Path` object, thanks to the `pydantic` autoconversion).

ifsbench/config_mixin.py
```diff
@@ -0,0 +1,36 @@
...
+__all__ = ['ConfigMixin', 'CLASSNAME', 'RESERVED_NAMES']
+
+CLASSNAME = 'classname'
+RESERVED_NAMES = [
+    CLASSNAME,
+]
+
+
+class ConfigMixin(BaseModel):
+
```
mlange05, line 15: It'd be good to have some docstring here to explain what this does and what it should be used for.

ifsbench/config_mixin.py
```diff
@@ -0,0 +1,36 @@
...
+    @classmethod
+    def from_config(
```
mlange05, line 17: Same here, all outward-facing API should have a docstring, I think.

ifsbench/config_mixin.py
```diff
@@ -0,0 +1,36 @@
...
+        cls, config: Dict[str, Union[str, float, int, bool, List, None]]
+    ) -> 'ConfigMixin':
+        return cls(**config)
+
+    def dump_config(
+        self, with_class: bool = False
+    ) -> Dict[str, Union[str, float, int, bool, List]]:
```
mlange05, line 24: Isn't `Union[str, float, int, bool, List]` not practically `Any`?

ifsbench/config_utils.py
```diff
@@ -0,0 +1,25 @@
...
+import yaml
+
+from ifsbench.config_mixin import CLASSNAME, ConfigMixin
+from ifsbench.data import ExtractHandler, NamelistHandler
+
+__all__ = ['read_yaml_config']
+
+_SUPPORTED_CONFIGS = {'ExtractHandler': ExtractHandler, 'NamelistHandler': NamelistHandler}
+
+def parse_config(config: Dict[str, Union[str, float, int, bool, List, None]]) -> Dict[str,ConfigMixin]:
```
mlange05, line 13: Docstring again, as public-facing API. 

The fact that this works as easy as this is amazing though! 🤩

ifsbench/data/namelisthandler.py
uhager: I changed the field types from str to Path. pydantic accepts str for Path but not vice versa. If we want to dump the config to yaml with a str, we can add 
```
def yaml_path_to_str(dumper,path):
    return dumper.represent_str(str(path))

yaml.add_multi_representer(Path, yaml_path_to_str)
```

mlange05 (changes requested):
Fantastic work, and very elegantly implemented! The only comments I have are nitpicks and a request for docstrings for all outer-facing API pieces. Will formally approve once some basic docstrings are in place.

Many thanks for the initiative and the fantastic implementation. 🙏 🚀 🤩

ifsbench/config_mixin.py
```diff
@@ -0,0 +1,36 @@
+from typing import Dict, List, Union
```
johannesbulin, line 1: Please add the same license header as in the other files!

ifsbench/config_utils.py
```diff
@@ -0,0 +1,25 @@
+import pathlib
```
johannesbulin, line 1: License header again.

tests/test_config_mixin.py
```diff
@@ -0,0 +1,91 @@
+from typing import Dict, List
```
johannesbulin, line 1: License header.

tests/test_config_utils.py
```diff
@@ -0,0 +1,87 @@
+import pathlib
```
johannesbulin, line 1: License header.

johannesbulin (commented):
Really happy about the changes! Thanks for the great idea with the mixin and the subsequent implementation. Once all license header have been added I am happy to approve this!

ifsbench/config_mixin.py
uhager: Done.

ifsbench/config_mixin.py
uhager: Done

ifsbench/config_mixin.py
uhager: In practice probably yes. But I wanted to make it explicit that these have to be types that can be supported when then dumping to and reading from yaml/toml/json.
After Johannes' requested change to make the NamelistHandler's input_/output_path Path objects, I actually added special handling to convert Path to str. I had originally intended to leave that to the yaml dumper, but then we'd have to reimplement for that for every supported format. 
So the output here should be whatever yaml can cleanly put into the config output and can then be read back and used again.

ifsbench/config_utils.py
uhager: This actually wasn't supposed to be public, renamed to _parse_config. read_yaml is the public function, I added a doc string there

ifsbench/config_mixin.py
uhager: Done.

tests/test_config_mixin.py
uhager: Done.

mlange05 (approved):
Looks great now. Thanks for the changes!

johannesbulin (approved):

PR #14, opened 2025-02-06
IFSBench: Install via pyproject.toml, absolute import paths and internalised entry points

mlange05:
This is a housekeeping PR that moves the installation meta information for pip to `pyproject.toml`, changes to a complete use of absolute imports throughout the package. With that in place, I've then also created a sub-package `cli` and moved the utilities, as well as the external entry point scripts into this internal sub-package. This then allows us to get rid of the `__main__ ` dunder calls and explicit shebangs. 

Integration note: Please check that the external-facing tools works as you'd expect and that the ifs-test CLI interfacing still works as one expects.


johannesbulin (approved):
Looks good so far - I've run some additional checks:
- CMake based install
- Pip install with optional packages
- Use `ifs-bench.py` scripts.

No issues so far so I am happy to get this merged.

pyproject.toml
```diff
@@ -1,11 +1,64 @@
...
+  "click",    # essential for CLI scripts
+  "f90nml",   # essential for namelist handling
+  "xarray",   # essential for data handling
+]
+
+[project.optional-dependencies]
+tests = [
+  "pytest",
+  "pytest-cov",
+  "coverage2clover",
```
reuterbal, line 32: This was just required to convert pytest output to XML for Bamboo to understand. I suggest we remove this now.

pyproject.toml
```diff
@@ -1,11 +1,64 @@
...
+  "pylint",
+  "pytest-pylint",
```
reuterbal, line 34: Similarly, this was required because it was the best way to route the pylint output via pytest to Bamboo. We're running this separately now, so no longer required.

reuterbal (changes requested):
Very nice, thanks for taking care of this!

Just a little pointer on two packages that are no longer required - not strictly related to this PR but it's a good enough opportunity to take care of this maybe?

PR #16, opened 2025-02-19
EnsembleStats class implementation

uhager:
The new class enables calculating various statistics based on a list of pandas DataFrames. The DataFrames are expected to all have the same structure and columns, with each one representing the results from a different ensemble member. EnsembleStats aggregates that data and provides summary statistics.

EnsembleStats can dump the raw data it receives into a json file, and can also be created from such a file.

Since EnsembleStats does not easily fit into the pydantic base model (we want to be able to create instances either from list of dataframes, or by reading a data file), ConfigMixin has been refactored with a purely abstract base class and then a pydantic implementation.

### DIFF
--- a/ifsbench/results/ensemble_stats.py
+++ b/ifsbench/results/ensemble_stats.py
@@ -1,0 +1,69 @@
+import json
+import pathlib
+import re
+from typing import Dict, List, Union
+
+import pandas as pd
+
+__all__ = ['EnsembleStats', 'ENSEMBLE_DATA_PATH']
+
+ENSEMBLE_DATA_PATH = 'ensemble_data'
+AVAILABLE_BASIC_STATS = ['min', 'max', 'mean', 'median', 'sum', 'std']
+
+_JSON_ORIENT = 'columns'
+
+
+def _percentile(nth):
+    def _qtile(data):
+        return data.quantile(nth / 100.0)
+
+    _qtile.__name__ = 'p{:02.0f}'.format(nth)
+    return _qtile
+
+
+class EnsembleStats:
+
+    def __init__(self, data: List[pd.DataFrame]):
+        self._raw_data = data
+        dfc = pd.concat(data)
+        self._group = dfc.groupby(dfc.index)
+
+    @classmethod
+    def from_data(cls, raw_data: List[pd.DataFrame]) -> 'EnsembleStats':
+        return cls(raw_data)
+
+    @classmethod
+    def from_config(cls, config: Dict[str, str]) -> 'EnsembleStats':
+        if not ENSEMBLE_DATA_PATH in config:
+            raise ValueError(f'missing config entry {ENSEMBLE_DATA_PATH}')
+        if len(config) > 1:
+            raise ValueError(
+                f'unexpected entries in config: {config}, expected only {ENSEMBLE_DATA_PATH}'
+            )
+        input_path = pathlib.Path(config[ENSEMBLE_DATA_PATH])
+        with open(input_path, 'r') as jin:
+            jdata = json.load(jin)
+        dfs = [pd.DataFrame.from_dict(json.loads(entry)) for entry in jdata]
+        return cls(dfs)
+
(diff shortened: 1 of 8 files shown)


ifsbench/results/ensemble_stats.py
```diff
@@ -0,0 +1,118 @@
...
+        def std(x):
+            # pandas uses sample standard deviation, we want population std.
+            # Using function rather than lambda for the correct column name.
+            return x.std(ddof=0)
+
+        if isinstance(stats, str):
+            stats = [stats]
+        to_request = []
+        for stat in stats:
+            percentile_check = re.match(r'[p,P](\d{1,2})$', stat)
```
johannesbulin, line 106: Should probably be `\d{1,3}` as we want to support `p100` (=maximum) as well.

uhager: Given that that is identical to max, I'm not sure we need to support that. If I allow 3 digits here, I'll also have to add a check that it's not >100 (numpy will fail, but then the failure message may be less clear since numpy uses quantiles between [0,1] so the error message will reflect that rather than that percentiles >100 are invalid).

But if you see a use-case for requiring P100, like some script that automatically generates a range of percentiles, I can add that.

johannesbulin: I agree that almost no-one would use `p100`. Still, the docstring specifies the percentile to be between 0 and 100 (inclusive, probably) and `numpy` supports this as well. I think it's better to support this and accept the few additional lines of code for error checking.

ifsbench/results/ensemble_stats.py
```diff
@@ -0,0 +1,118 @@
...
+            jdata = json.load(jin)
+        dfs = [pd.DataFrame.from_dict(json.loads(entry)) for entry in jdata]
+        es = cls(dfs)
+        es._data_file = config[ENSEMBLE_DATA_PATH]
+        return es
+
+    def dump_config(
+        self, with_class: bool = False
+    ) -> Dict[str, Union[str, float, int, bool, List]]:
+        if not self._data_file:
```
johannesbulin, line 72: I would prefer it if `dump_config` would not rely on having the frames dumped to a file. Instead I would like it to just use the `to_dict` functionality of `DataFrame` to create a nested dictionary, so something like
```
config['frames'] = [frame.to_dict() for frame in self._frames]
```
Rationale behind this:

- the file-based approach requires calling `dump_data_to_json` before `dump_config`, which is error prone, requires explicit paths and is in it's current form tied to JSON.
- if we have a result class that has several `EnsembleStats` attributes, it would be great if we could serialise such a result object into a single file, instead of having`result.json` file and `frame1.json` and `frame2.json` in the end.

uhager: Done.

ifsbench/results/ensemble_stats.py
```diff
@@ -0,0 +1,118 @@
...
+        js = [df.to_json(orient=_JSON_ORIENT) for df in self._raw_data]
+        with open(output_file, 'w', encoding='utf-8') as outf:
+            json.dump(js, outf)
+
+    def calc_stats(self, stats: Union[str, List[str]]) -> pd.DataFrame:
+        """Calculate statistics.
+
+        Desired statistics can be specified as a single string, e.g. 'mean', 'min',
+        or as a list of strings. Supported stats are given by AVAILABLE_BASIC_STATS;
+        in addition, percentiles between 0 and 100 can be specified as 'P10', 'p85', etc/
```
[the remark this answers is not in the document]
uhager: Done

ifsbench/results/ensemble_stats.py
uhager: Done

johannesbulin (approved):
Happy about the changes. Thanks for looking into the whole pandas environment and figuring out how to do this in a nice and efficient way!

reuterbal (approved):

PR #18, opened 2025-02-26
Add Sphinx documentation

reuterbal:
You can build the documentation like this:
```
pip install .[docs]
cd docs
make html
```

The piggy-backed GH Actions workflow will automatically upload this to https://sites.ecmwf.int/docs/ifsbench for any changes on the main branch.


.github/workflows/docs.yaml
```diff
@@ -0,0 +1,56 @@
...
+          python -m pip install --upgrade pip
+          python -m pip install -e .[docs]
+
+      - name: Build Sphinx documentation
+        run: |
+          cd docs
+          make html
+
+      - name: Deploy to sites
+        if: github.repository == 'ecmwf-ifs/ifsbench' && github.event_name != 'pull_request'
```
johannesbulin, line 39: In theory, we could trigger this action manually on a branch other than `main` and this would overwrite the central documentation. Might be good to prevent this somehow.

docs/source/conf.py
```diff
@@ -0,0 +1,122 @@
...
+# documentation root, use os.path.abspath to make it absolute, like shown here.
+#
+from importlib import metadata
+
+
+# -- Project information -----------------------------------------------------
+
+project = 'ifsbench'
+copyright = '2020- European Centre for Medium-Range Weather Forecasts (ECMWF)'
+author = 'Michael Lange, Balthasar Reuter, Johannes Bulin'
```
johannesbulin, line 29: Probably add Ulrike here as well.

johannesbulin (changes requested):
Thanks for the work. Looks good in my opinion. Once the potential manual-deploy-overwrites-central-documentation thing is fixed I'll happily approve this.

.github/workflows/docs.yaml
reuterbal: OK to simply remove the workflow_dispatch trigger?

johannesbulin: Fine with me!

reuterbal: Went for a ref_name check instead, to allows us to trigger the workflow manually if the upload failed for some reason

reuterbal:
Thanks for the feedback. I've rebased this and think it's ready for another look

johannesbulin (approved):

Issue #19, opened 2025-03-03
ifsbench pydantic usage


johannesbulin:
### Is your feature request related to a problem? Please describe.

For serialisation purposes we already use `pydantic` in a few places. It would probably be useful to use `pydantic` also to do input checks for us, instead of doing this by hand in various places.

PR #22, opened 2025-03-06
serialise env using pydantic (#19)

uhager:
EnvHandler and DefaultEnvPipeline inherit PydanticConfigMixin to enable serialisation. 

Several unrelated diffs due to reformatting.


ifsbench/data/namelisthandler.py
```diff
@@ -22,7 +22,7 @@
 __all__ = ['NamelistOverride', 'NamelistHandler', 'NamelistOperation']
 
 
-class NamelistOperation(Enum):
+class NamelistOperation(str, Enum):
```
johannesbulin, line 25: Just for my understanding: What is the purpose of making `str` one of the base-classes? Does this help `pydantic` with serialising? 
If possible, I think it would be better not to include `str` here, as this would (even if it's just a theoretical question) expose all `str` member functions here, which we probably don't want to have here.

johannesbulin (commented):
Thanks for all the changes (and for cleaning up some mistakes that I made). I am happy to approve this, once the `str/Enum` question is resolved.

ifsbench/data/namelisthandler.py
johannesbulin: Thanks! This also seems to be endorsed in the `pydantic` documentation: https://docs.pydantic.dev/2.1/usage/types/enums/. Happy with this now!

johannesbulin (approved):

ifsbench/env.py
```diff
@@ -171,35 +166,19 @@ class DefaultEnvPipeline(EnvPipeline):
...
+    handlers: List[EnvHandler] = Field(default_factory=list)
+    env_initial: Optional[Dict[str, Optional[str]]] = Field(default_factory=dict)
 
     def add(self, handler: Union[EnvHandler, List[EnvHandler]]) -> None:
         if isinstance(handler, EnvHandler):
-            self._handlers.append(handler)
+            self.handlers.append(handler)
         else:
-            self._handlers += handler
+            self.handlers += handler
```
reuterbal, line 176: [optional] This line is currently not covered by tests - which should be the scenario that `handler` is a list. It might be easy to capture this by adding a test parametrization which simply uses a 1-entry list. But that lack of coverage was there before, so not making this a requirement here

reuterbal (approved):
Many thanks, this looks fantastic! To me, this is ready to merge - I've pointed out a coverage omission which you could maybe resolve in one of the subsequent PRs?

PR #23, opened 2025-03-06
serialise RenameHandler using pydantic (#19).

uhager:


ifsbench/data/renamehandler.py
```diff
@@ -50,18 +54,14 @@ class RenameHandler(DataHandler):
...
-        repl: str,
-        mode: RenameMode = RenameMode.SYMLINK,
-    ):
-        if isinstance(pattern, re.Pattern):
-            self._pattern = pattern
-        else:
-            self._pattern = re.compile(pattern)
-        self._repl = str(repl)
-        self._mode = mode
+    pattern: str
```
johannesbulin, line 57: @reuterbal : Do I remember it correctly, that you would like to be able to pass a compiled `re.compile` pattern here as `pattern`? 
If so, @uhager , could we solve this by a `BeforeValidator` which just converts all `re.compile`-kind objects to `str` before the actual `pydantic` validation procedure hits?

ifsbench/config_mixin.py
```diff
@@ -88,7 +88,7 @@ def dump_config(
         Returns:
             Configuration that can be used to create instance.
         """
-        config = self.model_dump(exclude_none=True)
+        config = self.model_dump(exclude_none=True, round_trip=True)
```
johannesbulin, line 91: Just for my understanding, why do we need `round_trip` here? Which part of `RenameHandler` wouldn't serialise correctly without it?

johannesbulin (commented):
Happy about the changes so far - thanks for looking that deeply into `pydantic`. I would like to wait for @reuterbal  opinion regarding the `re.compile` thing, first.

ifsbench/config_mixin.py
uhager: This is to exclude computed_fields from the output. By default, the computed pattern would be added in dump_model, but we already have the pattern in string form and I would prefer the output to match the input.

ifsbench/data/renamehandler.py
reuterbal: I don't have a strict requirement for passing compiled patterns. And unless there is a use case, I'd say we don't need to introduce complexity until one arises.

The remark in my comment (https://github.com/ecmwf-ifs/ifsbench/pull/7#discussion_r1908824350) at the time was motivated by two considerations:
1. Let's compile the pattern in constructor instead of re-compiling it in each iteration of the file loop in the `execute` method. While this may ultimately not be measurable from a performance point of view thanks to the pattern cache of the `re` module, it made sense to me to have a static pattern pre-compiled instead of relying on that implicit functionality - and that is still the case here.
2. With that in mind, I simply thought "hey, then we can even pass compiled patterns" because it changed nothing at the time. It makes a difference now, so that point is moot :)

reuterbal (approved):
Very nice, many thanks! :shipit:

PR #24, opened 2025-03-06
Benchmark and validation classes

johannesbulin:
- Added a generic benchmark class that should cover our needs in a number of use cases.
- Added first validation routines for dataframes.

### DIFF
--- a/ifsbench/benchmark.py
+++ b/ifsbench/benchmark.py
@@ -22,10 +22,10 @@ from ifsbench.job import Job
 from ifsbench.launch import Launcher
 
 
-__all__ = ['DefaultScienceSetup', 'DefaultTechSetup', 'DefaultBenchmark']
+__all__ = ['ScienceSetup', 'TechSetup', 'Benchmark']
 
 @dataclass
-class DefaultScienceSetup:
+class ScienceSetup:
     """
     Generic (scientific) benchmark setup.
 
@@ -38,7 +38,7 @@ class DefaultScienceSetup:
     This kind of information is called the ``scientific`` setup as it describes
     _what_ gets benchmarked and _how_.
     Additional technical details that don't alter results (debug flags, additional
-    environment variables, etc.) should be specified in :class:`DefaultTechSetup`, instead.
+    environment variables, etc.) should be specified in :class:`TechSetup`, instead.
     """
     #: The application that gets benchmarked.
     application: Application
@@ -54,12 +54,12 @@ class DefaultScienceSetup:
     env_handlers:  List[EnvHandler] = field(default_factory=list)
 
 @dataclass
-class DefaultTechSetup:
+class TechSetup:
     """
     Additional technical details for benchmarks.
 
-    This dataclass can be used in combination with a :class:`DefaultScienceSetup` 
-    to setup a `DefaultBenchmark`. It encapsulates additional technical details 
+    This dataclass can be used in combination with a :class:`ScienceSetup` 
+    to setup a `Benchmark`. It encapsulates additional technical details 
     like debug flags, debug executables or performance-altering environment variables
     that do not change the results of the benchmark.
     """
@@ -78,7 +78,7 @@ class DefaultTechSetup:
     env_handlers:  List[EnvHandler] = field(default_factory=list)
 
 @dataclass
-class DefaultBenchmark:
+class Benchmark:
     """
     Generic benchmark implementation.
 
@@ -90,10 +90,10 @@ class DefaultBenchmark:
 
     #: The main (scientific) benchmark setup that describes _what_ gets benchmarked
     # and _which data_ is used.
-    science: DefaultScienceSetup
+    science: ScienceSetup
 
     #: Additional technical details that don't alter the results.
-    tech: Optional[DefaultTechSetup] = None
+    tech: Optional[TechSetup] = None
 
     def setup_rundir(self,
         run_dir: Path,
--- a/ifsbench/benchmark.py
+++ b/ifsbench/benchmark.py
@@ -5,6 +5,10 @@
 # granted to it by virtue of its status as an intergovernmental organisation
 # nor does it submit to any jurisdiction.
 
+"""
+Generic benchmark implementation.
+"""
+
 from dataclasses import dataclass, field
 import os
 from pathlib import Path
@@ -23,9 +27,18 @@ __all__ = ['DefaultScienceSetup', 'DefaultTechSetup', 'DefaultBenchmark']
 @dataclass
 class DefaultScienceSetup:
     """
-    Generic benchmark setup.
+    Generic (scientific) benchmark setup.
 
-    This dataclass holds generic information that is needed by benchmarks.
+    This dataclass encapsulates the information that is needed to describe a 
+    benchmark:
+      * The application that is benchmarked.
+      * The data pipeline.
+      * The environment variables that are needed.
+
+    This kind of information is called the ``scientific`` setup as it describes
+    _what_ gets benchmarked and _how_.
+    Additional technical details that don't alter results (debug flags, additional
+    environment variables, etc.) should be specified in :class:`DefaultTechSetup`, instead.
     """
     #: The application that gets benchmarked.
     application: Application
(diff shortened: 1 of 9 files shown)


ifsbench/benchmark.py
```diff
@@ -0,0 +1,177 @@
...
+    #: The data pipeline for the initial data setup.
+    data_handlers:  List[DataHandler] = field(default_factory=list)
+
+    #: Environment handlers that are used for the initial data setup.
+    env_handlers:  List[EnvHandler] = field(default_factory=list)
+
+    #: The application that gets benchmarked.
+    application: Optional[Application] = None
+
+    def merge(self, other):
```
johannesbulin, line 40: For some reason I didn't get typehints like `other: DefaultBenchmarkSetup` to work here. I don't know whether this is some `dataclass` oddity. If you have a solution for this, I'd be happy to fix this.

uhager: I don't think it's a dataclass issue, it's that you're using the class name inside the class that's defining the name so the type is not really known at this point.
Can you try `other: 'DefaultBenchmarkSetup'` . I think otherwise it should also be possible to import `Self` from typing_extensions and use that

ifsbench/validation/frame_close_validation.py
```diff
@@ -0,0 +1,77 @@
...
+from dataclasses import dataclass
+from typing import Any, List, Tuple
+
+import numpy
+from pandas import DataFrame
+
+from ifsbench.validation.frame_util import get_float_columns
+
+@dataclass
+class FrameCloseValidation:
```
uhager, line 17: Does this really need a class? It seems easier to just have the compare function and pass in rtol / atol with 0 as defaults.

ifsbench/validation/frame_util.py
```diff
@@ -0,0 +1,56 @@
+# (C) Copyright 2020- ECMWF.
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+
+__all__ = ['get_float_columns', 'get_int_columns']
+
+def get_float_columns(frame):
```
uhager, line 10: type annotations please (also for int_columns)

ifsbench/benchmark.py
```diff
@@ -0,0 +1,177 @@
...
+    """
+
+    #: The data pipeline for the initial data setup.
+    data_handlers:  List[DataHandler] = field(default_factory=list)
+
+    #: Environment handlers that are used for the initial data setup.
+    env_handlers:  List[EnvHandler] = field(default_factory=list)
+
+    #: The application that gets benchmarked.
+    application: Optional[Application] = None
```
uhager, line 38: I'm still trying to wrap my head around the design and which class contains what and who sets what where, but I think Application has DataHandlers and EnvHandlers. Should those be merged? What is the relation between the explicit handlers and the ones in the application?

ifsbench/benchmark.py
```diff
@@ -0,0 +1,177 @@
...
+
+        Returns
+        -------
+        The merged object.
+        """
+        data_handlers = self.data_handlers + other.data_handlers
+        env_handlers = self.env_handlers + other.env_handlers
+        application = self.application
+        if other.application:
+            application = other.application
```
uhager, line 61: why does 'other' take precedence if both are provided? I think in a member function I would expect that the other gets merged into self, so actually maybe this is better as a function outside the class as `def merge(first, second) -> DefaultBenchmarkSetup` to make it clear that it returns a new object rather than merging into the current.

ifsbench/benchmark.py
```diff
@@ -0,0 +1,177 @@
...
+            job = arch_result.job
+            if launcher is None:
+                launcher = arch_result.default_launcher
+            if launcher_flags is None:
+                launcher_flags = arch_result.default_launcher_flags
+
+        if launcher is None:
+            raise ValueError("No launcher was specified!")
+
+        cmd = self._setup.application.get_command(run_dir, job)
```
uhager, line 167: I may have missed it, but is there a check somewhere that the application is not None at this point? IIUC, you expectation is that  `setup` has an application and `tech_setup` does not (or vice versa?). If this is the case, maybe those need to be separate data types with one of them requiring an application and the other not even having that field.

ifsbench/validation/frame_util.py
johannesbulin: Yes, sorry. Will fix this.

ifsbench/benchmark.py
johannesbulin: The idea is to have the data setup as a two-step procedure:

1. Use the handlers in `DefaultBenchmarkSetup` to setup the input data initially.
2. Do some application-specific updates at runtime. IFS for example requires setting different namelist flags at runtime (run directory, number of processors, etc.) that can't be done at `setup_rundir`.

Ideally, `setup_rundir` is only executed once and if we re-run the benchmark, only the application-specific pipeline is executed.

ifsbench/benchmark.py
johannesbulin: No, you are right. I'll have a look.

ifsbench/benchmark.py
uhager: Ideally, this distinction should be clear (and even somehow enforced) in the code. At minimum, the data_handlers and env_handlers should be called something like initial_setup_data/env_handlers. It may also be worthwhile considering whether to make those a separate NewType or something using the same class implementation so it's absolutely clear that a handler is created as SetupDataHandler or RuntimeDataHandler and raises an error if used at the wrong time.

ifsbench/validation/frame_close_validation.py
uhager: slight preference to use the simpler pure function approach until we know a class is needed (that also then maybe inherits from something so has additional requirements on the API). But also ok to leave it since you have it.

ifsbench/tests/test_benchmark.py
```diff
@@ -0,0 +1,193 @@
...
+        TouchHandler(path='file1'),
+        TouchHandler(path='file2'),
+    ]
+
+    file_list = [Path('file1'), Path('file2')]
+
+    return DefaultBenchmarkSetup(
+        application = DefaultApplication(command=['pwd']),
+        env_handlers = env_handlers,
+        data_handlers = data_handlers
```
reuterbal, line 104: NB: PEP8 says not to use spaces around `=` when indicating keyword arguments

ifsbench/ifs.py
reuterbal, line 1: I guess the IFS-specific knowledge encoded in this module is lost for now. Should this not be removed in a separate PR where an "IFSApplication" or so is introduced? Or is that already defined in an ifs-test branch somewhere?

reuterbal (commented):
Many thanks, this looks sound and promising! I agree with @uhager's comments, many thanks for the diligent review!

ifsbench/ifs.py
johannesbulin: If I remember correctly, the idea was to move the IFS specific functionality out of `ifsbench` into `ifs-test`. This won't happen immediately, though. In this case, we can probably keep `ifs.py` in the repo until we do the actual move.

reuterbal: Yes, I agree that application-specific knowledge is best encoded next to the application. But I think it's better to first build this and remove redundant functionality subsequently.

ifsbench/benchmark.py
johannesbulin: Split `DefaultBenchmarkSetup` into two classes - `DefaultScienceSetup` and `DefaultTechSetup` and improved docstrings.

ifsbench/benchmark.py
```diff
@@ -0,0 +1,194 @@
...
+from ifsbench.data import DataHandler
+from ifsbench.env import EnvHandler, DefaultEnvPipeline
+from ifsbench.job import Job
+from ifsbench.launch import Launcher
+
+
+__all__ = ['DefaultScienceSetup', 'DefaultTechSetup', 'DefaultBenchmark']
+
+@dataclass
+class DefaultScienceSetup:
```
uhager, line 28: All the classes here are called DefaultSomething but don't implement a generic base class; is the plan to have other implementations? Should the base class already be defined then? Otherwise, can we remove the default for now? If other implementations are added it will anyway be a smaller refactor if all the references do not include the Default prefixed reference but the generic name.

ifsbench/benchmark.py
```diff
@@ -0,0 +1,194 @@
...
+__all__ = ['ScienceSetup', 'TechSetup', 'Benchmark']
+
+@dataclass
+class ScienceSetup:
+    """
+    Generic (scientific) benchmark setup.
+
+    This dataclass encapsulates the information that is needed to describe a 
+    benchmark:
+      * The application that is benchmarked.
```
uhager, line 34: Can you add something here along the lines of 'unless an application in TechSetup is provided' or '; will be overridden by TechSetup.application if provided'?

uhager (approved):

reuterbal (approved):
Great work, thanks!

PR #25, opened 2025-03-07
Serialise job (#19)

uhager:
Renamed copy to clone to avoid clash with pydantic.BaseModel.copy. Fixed bug in threads_per_node.


ifsbench/job.py
```diff
@@ -48,52 +48,51 @@ def threads_per_node(self):
         The number of logical cores per node (threads). This value is automatically derived
         from the above properties.
         """
-        return self.cores_per_node * self.threads_per_core
+        return self.cores_per_node() * self.threads_per_core
```
reuterbal, line 51: I think the API would be nicer if we declared `cores_per_node` (and for consistency) `threads_per_node` as properties. Then the look and feel of reading `Job.threads_per_core` and `Job.threads_per_node` is the same, even though one is derived from the other (which is an implementation detail that a user doesn't need to care about).

uhager: Done. I had already considered it, but decided to keep as is and just fix the bug in case there was a reason to have these as functions.

johannesbulin (approved):

reuterbal (approved):

PR #26, opened 2025-03-10
Serialise application

uhager:
pydantic cannot easily handle fields of abstract class DataHandler and I do not want to force all future datahandlers into pydantic so I do not want to make the base class inherit from PydanticConfigMixin. Used a cached_property instead to calculate the handlers and cache the result.
Created a separate dict to look up DataHandler classes from string and added an enum to avoid typos.


johannesbulin:
Thanks for looking into this! Supporting the whole `DataHandler` hierarchy is nasty. At the moment I'd lean towards pydantic-ifying `DataHandler` itself instead of doing the config trick in `Application`. My rationale behind this is

- I guess that all `DataHandler` subclasses should be something that could be represented by a dataclass/pydantic class. At least I can't think of anything that wouldn't fit in this scheme.
- Everything that implements it's own `Application` subclass may have to re-implement a similar approach to support `DataHandler` lists. 

@reuterbal : I'd be very happy if we could get a third opinion on this!

uhager:
> Supporting the whole `DataHandler` hierarchy is nasty. At the moment I'd lean towards pydantic-ifying `DataHandler` itself instead of doing the config trick in `Application`. My rationale behind this is
> 
> * I guess that all `DataHandler` subclasses should be something that could be represented by a dataclass/pydantic class. At least I can't think of anything that wouldn't fit in this scheme.

I just tried implementing it that way, and it's not as straight-forward as I had thought. See draft pull request https://github.com/ecmwf-ifs/ifsbench/pull/28/files
If that's preferred we can go forward with that, though it will limit our DataHandler implementations in future.

reuterbal:
Apologies for the slow response. I think I'm lost on the details here but conceptually I'd argue that a solution with a more flexible API is not a disadvantage as long as it retains the type checking properties that we strive for. 

I'm not quite following why Application subclasses need to re-implement the behaviour from this PR again?

So, while I agree that it's likely that pydantic-compatibility is possible with any DataHandler implementations that we can imagine, the implementation in #28 doesn't necessarily look like it improves the usability.

PR #28, opened 2025-03-11
Make DataHandler base class Pydantic config. This simplifies serialis…

uhager:
This simplifies serialiling DefaultApplication but limits flexibility of future DataHandler implementations.

### DIFF
--- a/ifsbench/application.py
+++ b/ifsbench/application.py
@@ -7,9 +7,13 @@
 
 from abc import ABC, abstractmethod
 from pathlib import Path
-from typing import List, Optional
+from typing import List, Union
 
-from ifsbench.data import DataHandler
+from pydantic import BeforeValidator, Field
+from typing_extensions import Annotated
+
+from ifsbench.config_mixin import PydanticConfigMixin
+from ifsbench.data import DataHandler, ExtractHandler, NamelistHandler, RenameHandler
 from ifsbench.env import EnvHandler
 from ifsbench.job import Job
 
@@ -106,7 +110,10 @@ class Application(ABC):
         return NotImplemented
 
 
-class DefaultApplication(Application):
+ListValidator = BeforeValidator(lambda in_value: in_value or [])
+
+
+class DefaultApplication(Application, PydanticConfigMixin):
     """
     Default application implementation.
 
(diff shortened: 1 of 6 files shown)


ifsbench/application.py
```diff
@@ -125,42 +132,27 @@ class DefaultApplication(Application):
...
-        else:
-            self._env_handlers = []
-
-        if library_paths:
-            self._library_paths = list(library_paths)
-        else:
-            self._library_paths = []
+    command: List[str]
+    data_handlers: Annotated[
+        List[Union[RenameHandler, ExtractHandler, NamelistHandler]], ListValidator
```
johannesbulin, line 137: Would `List[DataHandler]` work as well? Other packages that use `ifsbench` for testing/benchmarking might implement their own `DataHandler` classes which wouldn't be accepted here.

ifsbench/application.py
```diff
@@ -106,7 +110,10 @@ def get_command(self, run_dir: Path, job: Job) -> List[str]:
         return NotImplemented
 
 
-class DefaultApplication(Application):
+ListValidator = BeforeValidator(lambda in_value: in_value or [])
```
johannesbulin, line 113: Can't `pydantic` validate lists itself? Or do we have some edge-case here?

johannesbulin (commented):
Looks good. Thanks for also having a look at this approach! I personally prefer this approach over the one in PR 26.

ifsbench/application.py
uhager: It gets a bit messy, I think this works:
```
List[Union[tuple(DataHandler.__subclasses__())]]
``` 
though having a Callable as a List type seems wrong. Also, if any DataHandler implementation is also subclassed, we'd have to move that into a function to get the subclasses of the subclasses.

ifsbench/application.py
uhager: The reason I added this is that when testing we send explicit `datahandlers = None, env_handlers = None`. Because the parameter is set, default or default_factory is not called and we're left with None. I don't know if this is something that will happen when actually running; I can adjust the tests to omit None parameters from the config, then we can just use a `default_factory` without a special validator. But if None is a valid input value, then we need this validator.

uhager: Removed this after offline discussion: the user should not explicitly pass None, if no value is given the parameter should just not be added to the config.

johannesbulin (approved):
Really happy about this. We had additional discussions and what Ulrike implemented is more or less how the pydantic documentation recommends it: https://docs.pydantic.dev/latest/concepts/unions/#discriminated-unions-with-callable-discriminator.

reuterbal (approved):
Many thanks, this looks good to me and the implementation is very clean. I'll follow whatever decision you've taken regarding mandating Pydantic here.

One request, if I may: Can we add a small comment or so to the DataHandler base class about the role of the `handler_type`? That might help others in the future to implement new data handlers.

uhager:
> One request, if I may: Can we add a small comment or so to the DataHandler base class about the role of the `handler_type`? That might help others in the future to implement new data handlers.

Added comment.

PR #31, opened 2025-03-21
Launching an executable now returns stdout, stderr and return code.

johannesbulin:
Updated the actual executable-launch mechanism such that it returns the output (stdout, stderr) from the launch back to Python. Previously it was not possible to catch stdout/stderr during benchmarking

### DIFF
--- a/ifsbench/benchmark.py
+++ b/ifsbench/benchmark.py
@@ -151,7 +151,7 @@ class Benchmark:
         arch: Optional[Arch] = None,
         launcher: Optional[Launcher] = None,
         launcher_flags: Optional[List[str]] = None
-    ):
+    ) -> BenchmarkSummary:
         """
         Run the benchmark.
 
@@ -171,7 +171,8 @@ class Benchmark:
         Returns
         -------
         BenchmarkSummary:
-
+            BenchmarkSummary object that holds the output and the walltime
+            of the benchmark.
         """
 
         env_pipeline = DefaultEnvPipeline(handlers=self.science.env_handlers, env_initial=os.environ)
--- a/ifsbench/benchmark.py
+++ b/ifsbench/benchmark.py
@@ -12,10 +12,12 @@ Generic benchmark implementation.
 from dataclasses import dataclass, field
 import os
 from pathlib import Path
+from time import time
 from typing import List, Optional
 
 from ifsbench.application import Application
 from ifsbench.arch import Arch
+from ifsbench.config_mixin import PydanticConfigMixin
 from ifsbench.data import DataHandler
 from ifsbench.env import EnvHandler, DefaultEnvPipeline
 from ifsbench.job import Job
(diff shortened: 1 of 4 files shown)


ifsbench/benchmark.py
```diff
@@ -151,6 +167,11 @@ def run(self,
             A custom launcher to use. If None, the arch launcher is used.
         launcher_flags: list[str]
             Additional flags to be added to the launcher invocation.
+
+        Returns
```
uhager, line 171: Please update typing annotation of the function to include -> BenchmarkSummary

ifsbench/launch/launcher.py
```diff
@@ -44,6 +44,11 @@ class LaunchData:
     def launch(self):
```
uhager, line 44: also return type annotation

ifsbench/tests/test_util.py
```diff
@@ -46,3 +45,9 @@ def test_execute():
         text = 'abc\n' * 100
         execute(['echo', text], logfile=logfile)
         assert logfile.read_text() == text + '\n'
+
+        # Write to stderr
+        result = execute([sys.executable, '-c', 'import sys; print(\'foo bar\', file=sys.stderr)'], logfile=logfile)
+        assert logfile.read_text() == 'foo bar\n'
+        assert result.stdout == ''
```
uhager, line 52: is there a way to mock time() in pytest to explicitly set the values it returns for repeated calls? Then you could also test the time.

ifsbench/util.py
```diff
@@ -8,17 +8,32 @@
...
+    #: The captured standard output.
+    stdout: str
+
+    #: The captured standard error.
+    stderr: str
+
+    #: The return code of the execution.
+    exit_code: int
 
 def execute(command, **kwargs):
```
uhager, line 38: return type annotation

uhager (approved):

ifsbench/util.py
```diff
@@ -68,61 +82,82 @@ def execute(command, **kwargs):
     if dryrun:
         # Only print the environment when in dryrun mode.
         info('[ifsbench] Environment: ' + str(run_env))
-        return
+        return ExecuteResult(
```
reuterbal, line 85: The dryrun code path appears to be untested.

johannesbulin: Added a simple test for it and pushed it.

ifsbench/util.py
```diff
@@ -8,19 +8,35 @@
...
+    """
+    Results of an :func:`execute` invocation.
+    """
+
+    #: The captured standard output.
+    stdout: str
 
-def execute(command, **kwargs):
+    #: The captured standard error.
+    stderr: str
```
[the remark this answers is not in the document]
reuterbal: When this becomes a problem we could consider using a string ring-buffer implemented as a subclass of StringIO, and have the full dump only go to the logfile. Output of more than 1k lines or so to stdout/stderr is probably not that helpful anyway. That should now be easy to do after your changes.

reuterbal (approved):

PR #32, opened 2025-03-21
Added FetchHandler class

johannesbulin:
Added a `FetchHandler` data handler which allows us to download data from URLs. This would be necessary to support the current ecLand workflow.

### DIFF
--- a/ifsbench/data/fetchhandler.py
+++ b/ifsbench/data/fetchhandler.py
@@ -1,0 +1,62 @@
+# (C) Copyright 2020- ECMWF.
+# This software is licensed under the terms of the Apache Licence Version 2.0
+# which can be obtained at http://www.apache.org/licenses/LICENSE-2.0.
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+
+import pathlib
+from typing import Union
+import urllib.error
+import urllib.request
+
+from ifsbench.config_mixin import PydanticConfigMixin
+from ifsbench.data.datahandler import DataHandler
+from ifsbench.logging import debug
+
+__all__ = ['FetchHandler']
+
+class FetchHandler(DataHandler, PydanticConfigMixin):
+    """
+    Fetch a file from a given URL.
+    """
+
+    #: The source URL from where the file gets fetched.
+    source_url: str
+
+    #: Path where the file will be placed. If the path is relative, it will
+    #: be used relative to the run directory.
+    target_path: pathlib.Path
+
+    #: Whether the file will be fetched even if ``target_path`` exists already.
+    force: bool = False
+
+    #: If True, any errors encountered are ignored. Otherwise, a ``RuntimeError``
+    #: is raised on failure.
+    ignore_errors: bool = True
+
+    def execute(self, wdir: Union[str, pathlib.Path], **kwargs) -> None:
+        wdir = pathlib.Path(wdir)
+
+        target_path = self.target_path
+
+        if not self.target_path.is_absolute():
+            target_path = wdir/self.target_path
+
+        # Create the necessary parent folders.
+        target_path.parent.mkdir(parents=True, exist_ok=True)
+
+        if target_path.exists() and (not self.force):
+            debug(f"File {target_path} exists already and won't be fetched.")
+            return
+        if target_path.exists():
+            target_path.unlink()
+
+        debug(f"Download file from {self.source_url} to {target_path}.")
+
+        try:
+            urllib.request.urlretrieve(self.source_url, filename=target_path)
+        except urllib.error.URLError as ue:
+            debug("Fetching file failed: {ue}")
+            if not self.ignore_errors:
+                raise RuntimeError(str(ue)) from ue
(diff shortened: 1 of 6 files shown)


ifsbench/data/tests/test_fetchhandler.py
```diff
@@ -0,0 +1,126 @@
...
+    with (tmp_path/'hello.txt').open('w') as f:
+        f.write("Hello!")
+    with (tmp_path/'goodbye.txt').open('w') as f:
+        f.write("Goodbye!")
+
+    return tmp_path
+
+@pytest.mark.parametrize('force', [True, False])
+@pytest.mark.parametrize('ignore_errors', [True, False])
+def test_fetchhandler_force(run_dir, force, ignore_errors):
```
uhager, line 59: I think this is better to split into two. In code we want to reduce duplication, in tests it's better to be more verbose (DAMP not DRY).

ifsbench/data/fetchhandler.py
```diff
@@ -0,0 +1,62 @@
...
+import urllib.error
+import urllib.request
+
+from ifsbench.config_mixin import PydanticConfigMixin
+from ifsbench.data.datahandler import DataHandler
+from ifsbench.logging import debug
+
+__all__ = ['FetchHandler']
+
+class FetchHandler(DataHandler, PydanticConfigMixin):
```
uhager, line 19: Can this wait until my other change is in that makes DataHandler pydantic?

ifsbench/data/tests/test_fetchhandler.py
```diff
@@ -0,0 +1,126 @@
...
+import pytest
+
+from ifsbench.data import FetchHandler
+
+
+@pytest.mark.parametrize('source_url', ['file://home/some_file.tar.gz'])
+@pytest.mark.parametrize('target_path', [Path('subdir/target.tar.gz')])
+@pytest.mark.parametrize('force', [True, False])
+@pytest.mark.parametrize('ignore_errors', [True, False])
+def test_fetchhandler_serialise(source_url, target_path, force, ignore_errors):
```
uhager, line 23: For unit test I'd prefer a name that includes the function(s) being tested, i.e. something with _dump_config

ifsbench/data/tests/test_fetchhandler.py
```diff
@@ -0,0 +1,126 @@
...
+    with (run_dir/'goodbye.txt').open('r') as f:
+        result = f.read()
+
+    if force:
+        assert result == 'Hello!'
+    else:
+        assert result == 'Goodbye!'
+
+@pytest.mark.parametrize('ignore_errors', [True, False])
+def test_fetchhandler_ignore_errors(run_dir, ignore_errors):
```
uhager, line 86: Similar here: split into
test_fetchhandler_execute_ignore_errors
test_fetchhandler_execute_dont_ignore_errors

uhager (commented):
Can this wait until after my DataHandler PR is in?
Otherwise, mostly minor stuff about test naming, you can take it or leave it.

johannesbulin:
> Can this wait until after my DataHandler PR is in? Otherwise, mostly minor stuff about test naming, you can take it or leave it.

Yes, happy to wait for it. I would just rebase after the other PR is merged. Will have a look at the other non-pydantic issues that you've pointed out.

ifsbench/data/fetchhandler.py
```diff
@@ -0,0 +1,67 @@
...
+            debug(f"File {target_path} exists already and won't be fetched.")
+            return
+
+        if target_path.exists():
+            target_path.unlink()
+
+        debug(f"Download file from {self.source_url} to {target_path}.")
+
+        try:
+            urllib.request.urlretrieve(self.source_url, filename=target_path)
```
uhager, line 63: what kind of files are you expecting to fetch, roughly how large? Just wondering if we need to specify some custom timeout and/or retry options? In the same vein, I don't know what arguments urlretrieve takes, but can and should the kwargs be passed in here so the user can specify requests properties?

ifsbench/data/fetchhandler.py
```diff
@@ -0,0 +1,67 @@
...
+        target_path = self.target_path
+
+        if not self.target_path.is_absolute():
+            target_path = wdir/self.target_path
+
+        # Create the necessary parent folders.
+        target_path.parent.mkdir(parents=True, exist_ok=True)
+
+        if target_path.exists() and (not self.force):
+            debug(f"File {target_path} exists already and won't be fetched.")
```
uhager, line 54: I'm not sure that `debug` is sufficient here, I think I would upgrade this to info so it's definitely logged.

ifsbench/data/fetchhandler.py
johannesbulin: At the moment I guess that it's going to be a few hundred megabytes. If we need to pull much larger files we probably have to update this procedure. I am not 100% sure whether `urlretrieve` does a "streaming" download or if it puts the whole file content to memory first (which would be prohibitive for large files).

uhager: urlretrieve is marked as legacy in the documentation, would it be better to use urlopen together with shutil.copyfileobj? urlopen also can take optional parameters for timeout and SSL context, so if anything needs a certificate or key it can be supported.

reuterbal (approved):
Looks great. A remark about outdated API, otherwise gtg

ifsbench/data/fetchhandler.py
johannesbulin: Thanks for the suggestions! I've used Balthasar's suggestion and switched to `urlopen`. Seems to work quite nicely.

uhager (approved):

PR #35, opened 2025-03-27
Serialise arch

uhager:
Use lookup dict for launchers as we don't expect external launchers.

### DIFF
--- a/ifsbench/arch.py
+++ b/ifsbench/arch.py
@@ -134,7 +134,7 @@ class DefaultArch(Arch, PydanticConfigMixin):
         return self._launcher
 
     def get_cpu_configuration(self) -> CpuConfiguration:
-        return self._cpu_config
+        return self.cpu_config
 
     def process_job(self, job: Job, **kwargs) -> ArchResult:
         result = ArchResult()
--- a/ifsbench/arch.py
+++ b/ifsbench/arch.py
@@ -10,11 +10,14 @@ Architecture specifications
 """
 from abc import ABC, abstractmethod
 from dataclasses import dataclass, field
+from functools import cached_property
 from typing import List, Optional
 
+from ifsbench.config_mixin import PydanticConfigMixin
 from ifsbench.env import EnvHandler
 from ifsbench.job import CpuConfiguration, Job
 from ifsbench.launch.launcher import Launcher
+from ifsbench.launch.launcher_utils import LauncherLookup, Launchers
 
 __all__ = ['Arch', 'DefaultArch']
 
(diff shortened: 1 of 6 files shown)


ifsbench/arch.py
```diff
@@ -91,54 +94,53 @@ def process_job(self, job: Job, **kwargs):
...
-        partition: Optional[str] = None,
-    ):
-        """
-        Default architecture that can be used for various systems.
+    """
+    Default architecture that can be used for various systems.
+
+    Parameters
+    ----------
+    launcher : Launcher
```
johannesbulin, line 104: As you've taken away the constructor, could you move the parts of the docstring that describe the parameters directly to the variable definitions (example below). If I remember, this is necessary for Sphinx being able to pick this up correctly.
```suggestion
    #: The default launcher that is used on this system.
    launcher : Launcher
    
    #: The hardware setup of the system.
    cpu_config: CpuConfiguration
    
    #: If set to True, the following attributes in result.job are
    #: calculated and set explicitly, using ``cpu_config``:
    #:      * tasks
    #:      * nodes
    #:      * tasks_per_node
    #:  If not, these values will stay None, if not specified.
    set_explicit: bool = False
    
    ...
```

ifsbench/arch.py
```diff
@@ -91,54 +94,53 @@ def process_job(self, job: Job, **kwargs):
...
+            * tasks_per_node
+        If not, these values will stay None, if not specified.
+    account : str
+        The account that is passed to the launcher.
+    partition : str
+        The partition that will be passed to the launcher.
+    """
+
+
+    launcher: Launchers
```
johannesbulin, line 122: Not sure whether this approach will work in the future. While the current launcher classes can be instantiated by just their name, we might have stuff like
```
class ValgrindLauncher(PydanticConfigMixin):
  valgrind_options: List[str]
```
i.e. `Launcher` subclasses that actually have constructor arguments. In this case, it wouldn't be possible to these launchers in `DefaultArch` (yes, I know that valgrind is probably not going to be the default launcher anywhere, but we may have other such classes).

Can we do the launchers what we did for the data handlers? Have a `launcher_type` literal for each class so pydantic handles this automatically? Or is there something that may cause issues there?

johannesbulin (changes requested):

ifsbench/arch.py
uhager: I had assumed Launchers wouldn't need constructors; if that is required then this approach doesn't work. I created https://github.com/ecmwf-ifs/ifsbench/pull/36 to keep the Launcher refactoring separate, makes it easier to change it as needed. PTAL. 
Will update this PR once we have agreed on how to handle Launcher.

PR #39, opened 2025-04-02
Netcdf support

uhager:
Refactored gribfile to pull out the statistics calculations and have separate file readers for different file types.
The statistics calculations are now more flexible in terms of stat names and dimensions to average over as they will be used by a wider variety of data formats.
Made the new stats class inherit ConfigMixin so it can be added to yaml configs.


uhager:
Apologies for the large PR, getting git to apply the history from one file to the copy seems to be rather complicated, so all the stats code looks new even though it's unchanged.

ifsbench/data_file_reader.py
```diff
@@ -0,0 +1,26 @@
...
+
+class DataFileReader(ABC):
+
+    @classmethod
+    @abstractmethod
+    def read_data(cls, input_path: str) -> List[xr.Dataset]:
+        raise NotImplementedError()
+
+
+class NetcdfFileReader(DataFileReader):
```
johannesbulin, line 17: Could you either put `NetcdfFileReader` into its own file (as the `GribFileReader`) or put all readers into `data_file_reader.py`? Preferably the former.

ifsbench/data_file_stats.py
```diff
@@ -0,0 +1,131 @@
...
+    DataFileType.GRIB: GribFileReader,
+    DataFileType.NETCDF: NetcdfFileReader,
+}
+
+
+class DataFileStats(PydanticConfigMixin):
+
+    input_path: Path
+    filetype: Optional[DataFileType] = None
+    stat_dims: Set[str] = _DEFAULT_STAT_DIMS
```
johannesbulin, line 54: Are the dims in an `xarray` always string values? Sorry, I don't know much about `xarray` but was wondering if they might be ints (or something else) as well.

ifsbench/data_file_stats.py
```diff
@@ -0,0 +1,131 @@
...
+_DEFAULT_STAT_NAMES = ['mean', 'min', 'max', 'p5', 'p10', 'p90', 'p95']
+
+# Dimension name to add to statistics datasets and use for merging.
+# This will be the column name in the dataframe with values from
+# _STAT_NAMES
+_STAT_DIM_NAME = 'stat'
+
+
+class DataFileType(str, Enum):
+    GRIB = 'grib'
```
johannesbulin, line 40: Please add docstrings/sphinx annotations (also for `DataFielStats`).

ifsbench/data_file_stats.py
```diff
@@ -0,0 +1,131 @@
...
+
+_reader_from_file_type = {
+    DataFileType.GRIB: GribFileReader,
+    DataFileType.NETCDF: NetcdfFileReader,
+}
+
+
+class DataFileStats(PydanticConfigMixin):
+
+    input_path: Path
```
johannesbulin, line 52: As all readers return `xarray` objects, wouldn't it make sense to take an `xarray` here, instead of `input_path` and `filetype`? This would allow us to do the same statistics stuff with arbitrary `xarray` objects, even if they are not in a file.

pyproject.toml
```diff
@@ -22,6 +22,7 @@ dependencies = [
   "pydantic", # essential for typing and configs
   "click",    # essential for CLI scripts
   "f90nml",   # essential for namelist handling
+  "netcdf4",  # essential for reading netcdf files
```
johannesbulin, line 25: Should this become a main dependency or an optional one. I am fine with having it as a main dependency _if_ this works on all systems and doesn't cause weird issues like the `pygrib` module.

uhager: I think netcdf is a lot better supported and more widely known. There are no issues with any of the github actions, so I think it's ok to just use it. At some point we have to add some library that is able to read in some data, and netcdf is probably the easiest (see the one-line implementation for opening it).

ifsbench/data_file_stats.py
uhager: I think I'd rather not. For data can get rather large, and I don't think we want it serialised when dumping the config.  If we really come across a use-case, we can consider some alternative where we we make it impossible to accidentally end up outputting the full result of a forecast into a yaml file.

ifsbench/data_file_stats.py
uhager: yes. I think a lot of things would break if they were suddenly ints.

ifsbench/data_file_reader.py
```diff
@@ -0,0 +1,26 @@
...
+import xarray as xr
+
+__all__ = ['DataFileReader', 'NetcdfFileReader']
+
+
+class DataFileReader(ABC):
+
+    @classmethod
+    @abstractmethod
+    def read_data(cls, input_path: str) -> List[xr.Dataset]:
```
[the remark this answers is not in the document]
uhager: Done.

ifsbench/data_file_reader.py
```diff
@@ -0,0 +1,29 @@
...
+
+from abc import ABC, abstractmethod
+from typing import List
+
+import xarray as xr
+
+__all__ = ['DataFileReader']
+
+
+class DataFileReader(ABC):
```
johannesbulin, line 16: Sorry, missed it the first time. The class itself should also get a docstring and a short description what it is about.

uhager: Done.

johannesbulin (approved):

Issue #40, opened 2025-04-03
Can't read DefaultArch from YAML file


johannesbulin:
### What happened?

Deserialising/Validating a `DefaultArch` object fails if one tries to do 
```
DefaultArch.from_config(config_dict)
```
The error message is
```
  File "/Users/johannes/Desktop/launcher_stuff/dump_recover.py", line 14, in <module>
    DefaultArch.from_config(config)
  File "/Users/johannes/Desktop/launcher_stuff/venv/lib/python3.11/site-packages/ifsbench/config_mixin.py", line 78, in from_config
    return cls(**config)
           ^^^^^^^^^^^^^
  File "/Users/johannes/Desktop/launcher_stuff/venv/lib/python3.11/site-packages/pydantic/main.py", line 243, in __init__
    validated_self = self.__pydantic_validator__.validate_python(data, self_instance=self)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: Can't instantiate abstract class Launcher with abstract method prepare
```

### What are the steps to reproduce the bug?

Download these files, rename `config.txt` to `config.yaml` and the other two files to `.py` files. Run
* `from_yaml.py` to recover `DefaultArch` from a YAML file.
* `dump_recover.py` to run a `dump_config/from_config` chain.

[config.txt](https://github.com/user-attachments/files/19583898/config.txt)
[dump_recover.txt](https://github.com/user-attachments/files/19583897/dump_recover.txt)
[from_yaml.txt](https://github.com/user-attachments/files/19583896/from_yaml.txt)

### Version

main

### Platform (OS and architecture)

all

### Relevant log output

```shell

```

PR #43, opened 2025-04-03
Fix subclass resolution when deserialising (#40).

uhager:
Found on https://github.com/pydantic/pydantic/issues/7366#issuecomment-1711827295
This should finally allow subclassing with serialisation and deserialisation and adding subclasses at runtime that can be used as their base type in other pydantic classes.

Also clean up redundant Field(default='xyz') statements in Launcher and DataHandler subclasses.

### DIFF
--- a/ifsbench/data/datahandler.py
+++ b/ifsbench/data/datahandler.py
@@ -6,9 +6,13 @@
 # nor does it submit to any jurisdiction.
 
 from abc import abstractmethod
-from typing import Union
-
 from pathlib import Path
+
+from typing import Any, ClassVar, Dict, Type, Union
+
+from pydantic import model_validator, TypeAdapter, Field
+from pydantic_core.core_schema import ValidatorFunctionWrapHandler
+from typing_extensions import Annotated
 
 from ifsbench.config_mixin import PydanticConfigMixin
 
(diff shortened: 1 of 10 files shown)


ifsbench/data/datahandler.py
```diff
@@ -28,6 +32,28 @@ class DataHandler(PydanticConfigMixin):
...
+    def _parse_into_subclass(
+        cls, v: Any, handler: ValidatorFunctionWrapHandler
+    ) -> 'DataHandler':
+        if cls is DataHandler:
+            return DataHandler._discriminating_type_adapter.validate_python(v)
+        return handler(v)
+
+    @classmethod
+    def __pydantic_init_subclass__(cls, **kwargs):
+        DataHandler._subclasses[cls.model_fields['handler_type'].default] = cls
```
johannesbulin, line 49: Do we need a call to `super().__pydantic_init_subclass` here?

ifsbench/launch/launcher.py
```diff
@@ -75,16 +79,27 @@ class Launcher(PydanticConfigMixin):
...
+        cls, v: Any, handler: ValidatorFunctionWrapHandler
+    ) -> 'Launcher':
+        if cls is Launcher:
+            return Launcher._discriminating_type_adapter.validate_python(v)
+        return handler(v)
 
-        return value
+    @classmethod
+    def __pydantic_init_subclass__(cls, **kwargs):
+        Launcher._subclasses[cls.model_fields['launcher_type'].default] = cls
```
johannesbulin, line 96: As above, do we have to call the superclass function, as well?

johannesbulin (commented):
I like this approach - should solve most of our remaining issues with serialisation - still have to test deserialisation, as well.

At the moment, each base class has to implement these three functions, which is a lot of code duplication and additional overhead. Do you think it would work to 

- either put the logic directoy into `PydanticConfigMixin`
- create a class decorator (`@pydantic_baseclass` or something like this) which just adds the functionality to a class?

ifsbench/data/datahandler.py
uhager: super() just calls pass: https://github.com/pydantic/pydantic/blob/main/pydantic/main.py#L830

uhager:
> I like this approach - should solve most of our remaining issues with serialisation - still have to test deserialisation, as well.
> 
> At the moment, each base class has to implement these three functions, which is a lot of code duplication and additional overhead. Do you think it would work to
> 
> * either put the logic directly into `PydanticConfigMixin`
> * create a class decorator (`@pydantic_baseclass` or something like this) which just adds the functionality to a class?

I am planning to try and put this in the baseclass somehow with a str field that specifies the discriminator field value. But I'm not sure yet if that will work so to unblock your work, I left the duplication for now and will try to find a cleaner solution in a follow-up PR (if I do find a cleaner solution).

johannesbulin (approved):

PR #44, opened 2025-04-10
Added pydantic-annotation-wrapper for pandas.DataFrame.

johannesbulin:
Pydantic can't serialise `pandas.DataFrame` objects out-of-the-box which is annoying if we have result classes like
```
class SomeResult(PydanticConfigMixin):
  some_result: pandas.DataFrame
```
I've followed the examples in https://docs.pydantic.dev/latest/concepts/types/#handling-third-party-types to create a custom annotated-wrapper for `pandas.DataFrame` that makes it easier to use `pandas.DataFrame` in combination with `pydantic`. The previous example can easily be converted to
```
from ifsbench import PydanticDataFrame

class SomeResult(PydanticConfigMixin):
  some_result: PydanticDataFrame
```
and should work.


ifsbench/pydantic.py
uhager, line 1: nit: maybe pydantic_utils? pydantic_types? Depends a bit on what else you think might be added.

ifsbench/pydantic.py
```diff
@@ -0,0 +1,76 @@
...
+                    # Support creation/validation from other DataFrame objects
+                    # and from dictionaries.
+                    core_schema.is_instance_schema(DataFrame),
+                    from_dict_schema,
+                ]
+            ),
+            serialization=core_schema.plain_serializer_function_ser_schema(
+                # Serialise a frame. We have to use `orient=split` here, as
+                # other orient-values may lead to a reordering of columns.
+                lambda frame: frame.to_dict(orient='split')
```
uhager, line 69: In EnsembleStats (which I might look into rewriting using PydanticDataFrame) I use to_dict without specifying orient, but then I also use from_dict rather than the direct DataFram(**dict). Should we make this consistent? I don't remember whether I properly checked this back then (shame on me) what it looks like, I think I may have printed it to see it looks readable when output to yaml.

Maybe using the explicit `from_dict` in `validate_from_dict` and using `orient='tight'` for all `to_dict` and `from_dict` (here and EnsembleStats and any other df<->dict conversion) would work and be least error-prone? Haven't tested though, and I'm also ok with leaving it as is if you think it's cleanest.

uhager (commented):
Fancy! Very nice!

ifsbench/tests/test_pydantic_dataframe.py
```diff
@@ -0,0 +1,74 @@
...
+    """
+    obj = _PydanticDataFrameTest(
+        frame=default_frames[0],
+        frame_dict={'name': default_frames[1]},
+        frame_list=[default_frames[2]]
+    )
+
+    config = obj.dump_config()
+
+    copy = _PydanticDataFrameTest.from_config(config)
```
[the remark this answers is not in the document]
johannesbulin: Makes sense. Will do it.

ifsbench/pydantic.py
johannesbulin: Switched to `orient='tight'` and added more tests for this.

ifsbench/tests/test_pydantic_dataframe.py
```diff
@@ -72,3 +101,73 @@ def test_pydantic_data_frame_recover(default_frames):
...
+    )
+
+    config = obj.dump_config()
+
+    path = tmp_path/'file.json'
+
+    with path.open('w') as f:
+        json.dump(config, f)
+    with path.open('r') as f:
+        data = json.load(f)
```
uhager, line 128: This really just tests that json can read what it writes. You already tested above the the dumped config is sensible and easily human-readable, so I think you can skip this here.

uhager (approved):

PR #49, opened 2025-04-28
Fixed wrong flags when using the MpirunLauncher.

johannesbulin:
Apparently, `mpirun` doesn't use double-dashed options like most other tools and doesn't understand `mpirun --n=2` but only `mpirun --n 2`. I've fixed this by switching to single-dash options and removing the `=`.

### DIFF
--- a/ifsbench/launch/mpirunlauncher.py
+++ b/ifsbench/launch/mpirunlauncher.py
@@ -24,10 +24,10 @@ class MpirunLauncher(Launcher):
     launcher_type: Literal['MpirunLauncher'] = 'MpirunLauncher'
 
     _job_options_map = {
-        'tasks': '--n={}',
-        'tasks_per_node': '--npernode={}',
-        'tasks_per_socket': '--npersocket={}',
-        'cpus_per_task': '--cpus-per-proc={}',
+        'tasks': '-np {}',
+        'tasks_per_node': '-npernode {}',
+        'tasks_per_socket': '-npersocket {}',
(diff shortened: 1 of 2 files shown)


reuterbal (commented):
I agree with Ricki's comment. Moreover, the OpenMPI docs suggest that they may be undergoing some renaming/refactoring there at the moment: https://docs.open-mpi.org/en/v5.0.x/man-openmpi/man1/mpirun.1.html#options-old-hard-coded-content-mdash-to-be-audited

Naturally, this may pose problems when we have to support multiple versions that may not agree on a nomenclature. I'm not sure if there's a "guaranteed to be compatible with old and new versions" set of options, but pointing this out since it may happen that we otherwise end up fixing this again in relatively short time.

Particularly, since I can't find the single-dash option in the above link at all.

johannesbulin:
Thanks for the feedback, both of you. I agree that the content of the PR and its content is confusing, to put it mildly. I had a look at the OpenMPI link that Balthasar sent and will update this PR to use

- `-n <nprocess>` to set the number of processes
-  double-dash flags for everything else (`--npernode/--npersocket/--cpus-per-proc`)

Rationale behind this: This seems to be supported by OpenMPI 5.0 as well as older versions and will hopefully cause the least amount of trouble. Single options like `-npernode` are not mentioned in the OpenMP 5.0 documentation, therefore it's questionable that this works. It also states that `-n` is the preferred option to set the number of processes.

reuterbal (approved):

uhager (approved):

PR #50, opened 2025-05-06
Fixed problem with mpirun flags not being split into separate strings.

johannesbulin:
Unfortunately, #49 introduced new problems with `mpirun`, as flags like `-n 4` were not split up properly, resulting in commands like `['mpirun', '-n 4', 'executable']` being passed to the `subprocess` module.
This caused problems when actually launching the executable.

I've resolved this by splitting up things like `-n 4` into `['-n', '4']` which seems to work properly.

### DIFF
--- a/ifsbench/launch/mpirunlauncher.py
+++ b/ifsbench/launch/mpirunlauncher.py
@@ -79,7 +79,10 @@ class MpirunLauncher(Launcher):
             value = getattr(job, attr, None)
 
             if value is not None:
-                flags += [option.format(value)]
+                # Split the resulting split up if it contains spaces. Otherwise
+                # we might end up with a command like ['mpirun', '-n 4', 'program']
+                # which causes errors. We want ['mpirun', '-n', '4', 'program'].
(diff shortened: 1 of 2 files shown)


ifsbench/launch/mpirunlauncher.py
```diff
@@ -48,11 +48,10 @@ def _get_distribution_options(self, job: Job) -> List[str]:
         do_nothing = [
             CpuDistribution.DISTRIBUTE_DEFAULT,
             CpuDistribution.DISTRIBUTE_USER,
+            None
```
uhager, line 51: Is this change and the hasattr check removal below related to the flag splitting? why is this needed? Should this be a separate PR, and maybe add a CpuDistribution.None field? What happens if it's none?

johannesbulin: Yes, sorry - I've removed this now. You were correct that this is not connected to the flag splitting. Will put this in a later PR.

uhager (approved):

PR #52, opened 2025-06-13
[Hotfix] Define GPU requirements as per-node property rather than per-task

reuterbal:
This is necessary because the recent update of SLURM on the Atos Slurm controller introduced an incompatibility with the older slurm clients on login nodes, which no longer allows to specify --gpus-per-task. As a consequence, job submission is currently refused.

In the meantime, oversubscribing GPUs has also be found to be beneficial, thus a task-to-gpu mapping is no longer a good concept anyway. Therefore, this PR flips the GPU resource specification to a per-node resource. I have tried to retain most of the previous behaviour as possible (and in particular behaviour in our use cases is the same as before) but some edge cases have now (intentionally) a different resource allocation to allow for GPU oversubscription.

This PR is filed against a release-branch that is identical to 0.1.2 and does not include all of the developments that occured in the meantime. This is so that we can tag these changes as 0.1.3 while continuing work towards version 0.2 on main.


awnawab:
```
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 550.163.01             Driver Version: 550.163.01     CUDA Version: 12.4     |
|-----------------------------------------+------------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id          Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |           Memory-Usage | GPU-Util  Compute M. |
|                                         |                        |               MIG M. |
|=========================================+========================+======================|
|   0  NVIDIA A100-SXM4-40GB          On  |   00000000:03:00.0 Off |                    0 |
| N/A   41C    P0             72W /  400W |   39262MiB /  40960MiB |      0%      Default |
|                                         |                        |             Disabled |
+-----------------------------------------+------------------------+----------------------+
|   1  NVIDIA A100-SXM4-40GB          On  |   00000000:44:00.0 Off |                    0 |
| N/A   39C    P0             55W /  400W |       5MiB /  40960MiB |      0%      Default |
|                                         |                        |             Disabled |
+-----------------------------------------+------------------------+----------------------+
|   2  NVIDIA A100-SXM4-40GB          On  |   00000000:84:00.0 Off |                    0 |
| N/A   39C    P0             54W /  400W |       5MiB /  40960MiB |      0%      Default |
|                                         |                        |             Disabled |
+-----------------------------------------+------------------------+----------------------+
|   3  NVIDIA A100-SXM4-40GB          On  |   00000000:C4:00.0 Off |                    0 |
| N/A   39C    P0             53W /  400W |       5MiB /  40960MiB |      0%      Default |
|                                         |                        |             Disabled |
+-----------------------------------------+------------------------+----------------------+
```

Many many thanks for this 🙏 This does allow you to indeed run, but the binding doesn't seem to be set correctly ⬆️ , which makes it hard to run tco399 tests.

reuterbal:
Ah, indeed, I feared this might be the case. I have an idea for a workaround, hang on...

mlange05 (commented):
I've highlighted one particular change I needed to do locally, but otherwise I can confirm that this behaves perfectly as expected for dual-node GPU runs.

johannesbulin (approved):
Code looks fine. I've also created a follow-up issue to get this into main (#53).

johannesbulin (approved):

PR #54, opened 2025-06-17
[Hotfix] Atos_ac GPU affinity

reuterbal:
A small hotfix to #52: Turns out the affinity mapping of CPUs to GPUs is actually not linear on Atos_ac:

```
        GPU0    GPU1    GPU2    GPU3    NIC0    NIC1    CPU Affinity    NUMA Affinity   GPU NUMA ID
GPU0     X      NV4     NV4     NV4     SYS     SYS     32-63,160-191   1               N/A
GPU1    NV4      X      NV4     NV4     PIX     SYS     0-31,128-159    0               N/A
GPU2    NV4     NV4      X      NV4     SYS     PIX     96-127,224-255  3               N/A
GPU3    NV4     NV4     NV4      X      SYS     SYS     64-95,192-223   2               N/A
NIC0    SYS     PIX     SYS     SYS      X      SYS
NIC1    SYS     SYS     PIX     SYS     SYS      X

Legend:

  X    = Self
  SYS  = Connection traversing PCIe as well as the SMP interconnect between NUMA nodes (e.g., QPI/UPI)
  NODE = Connection traversing PCIe as well as the interconnect between PCIe Host Bridges within a NUMA node
  PHB  = Connection traversing PCIe as well as a PCIe Host Bridge (typically the CPU)
  PXB  = Connection traversing multiple PCIe bridges (without traversing the PCIe Host Bridge)
  PIX  = Connection traversing at most a single PCIe bridge
  NV#  = Connection traversing a bonded set of # NVLinks

NIC Legend:

  NIC0: mlx5_0
  NIC1: mlx5_1
```

Thus we need to shuffle around the visible devices per rank accordingly.

### DIFF
--- a/ifsbench/arch.py
+++ b/ifsbench/arch.py
@@ -318,23 +318,33 @@ class Atos(Arch):
 
             # To ensure we correctly bind GPUs to CPUs we write a little
             # wrapper script that maps the correct CUDA_VISIBLE_DEVICES
-            # id to each rank locally per node
+            # id to each rank locally per node.
+            # Because the mapping from NUMA domain to GPU is
+            # [1->0, 0->1, 3->2, 2->3], we need to apply some reordering here
 
-            if tasks_per_node <= gpus_per_node:
-                # With less or equal tasks per node than GPUs available,
-                # we can map one GPU to each task, in ascending order
-                # of local rank number per node
-                gpu_mapping_str = '${SLURM_LOCALID}'
-            else:
-                # With more ranks than GPUs available per node,
-                # we map subsequent rank ids on a node to the same
-                # GPU - effectively a compact allocation strategy
-                ranks_per_gpu = (tasks_per_node - 1) // gpus_per_node + 1
-                gpu_mapping_str = f'$((SLURM_LOCALID / {ranks_per_gpu}))'
+            # The GPU that is closest to each core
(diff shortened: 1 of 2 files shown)


johannesbulin:
Testing this with
```
cpus_per_task=8
threads_per_core=1
tasks_per_node=4
```
yields
```
CUDA_VISIBLE_DEVICES=1
```
for all four tasks per node. Is this intended or should each task access its own GPU?

Might an alternative be that we bind the tasks to the NUMA regions so that we don't have to do the GPU handling by hand?

reuterbal:
Thanks, I did consider this but decided to leave it like this for the moment.

It is unfortunately expected and it would require a bit more logic to handle this differently. The reason is that Slurm will in this instance allocate the 8 tasks to the 8x4=32 cores of the first NUMA domain (i.e., core 0-31), which are next to GPU 1. So the question is whether the expected behaviour in this situation is to allocate different GPUs that are not in the correct affinity domain, or allocate all tasks to the GPU(s) in the correct affinity domain.

In my opinion, we would need more user input in this situation to decide what the intended distribution is but I'm open to other suggestions. Or, ideally, the SLURM setup would be improved to not make this necessary in the first place...

johannesbulin (approved):
Thanks @reuterbal . In this case it is fine and I can't think of a nicer solution.

mlange05 (approved):
Looks good to me. I can confirm that there's no change to the dual-node GPU runs with this, so all good from my side. :shipit:

PR #56, opened 2025-07-01
Simplify serialisation of abstract classes and switch to pydantic completely

johannesbulin:
I made some draft changes to a couple of things where I would like to get your opinion on this.

### Use pydantic exclusively for serialisation

Until now, we were using `pydantic` (as `PydanticConfigMixin`) alongside the non-pydantic `ConfigMixin` to serialise objects. That's quite flexible but once we start mixing pydantic and non-pydantic classes we have to write quite a bit of glue-code. In my opinion, we can serialise everything that we need with `pydantic` so it might be easier to only use `pydantic`.

The only class that's strongly affected by this is the `EnsembleStats` class. I've gutted it quite a bit and pydantic-ified it. Please check whether this is acceptable as well

### Changed serialisation of pydantic class hierarchies

We had some issues with serialising hierarchies of pydantic classes where we had to add pydantic code to each "base class", as well as discriminator fields to each subclass. I've tried to solve it with an `AbstractDataClass` parent class that should do all these things automatically. The tests seem to work but there might be some corner cases where this fails.

### DIFF
--- a/ifsbench/arch.py
+++ b/ifsbench/arch.py
@@ -9,7 +9,7 @@
 Architecture specifications
 """
 from abc import ABC, abstractmethod
-from typing import List, Optional
+from typing import List
 
 from ifsbench.serialisation_mixin import SubclassableSerialisationMixin, SerialisationMixin
 from ifsbench.env import EnvHandler
--- a/ifsbench/data/namelisthandler.py
+++ b/ifsbench/data/namelisthandler.py
@@ -186,8 +186,6 @@ class NamelistSanitiseHandler(DataHandler):
     using different resolution strategies.
     """
 
-    handler_type: Literal['NamelistSanitiseHandler'] = 'NamelistSanitiseHandler'
-
     #: The path to the namelist that will be modified. If a relative path is
     #: given, this will be relative to the ``wdir`` argument in :meth:`execute`.
     input_path: pathlib.Path
--- a/ifsbench/data/tests/test_namelisthandler.py
+++ b/ifsbench/data/tests/test_namelisthandler.py
@@ -522,7 +522,6 @@ def test_namelisthandlersanitise_dump_config(mode):
     nh = NamelistSanitiseHandler.from_config(config)
 
     expected = {
-        'handler_type': NamelistSanitiseHandler.__name__,
         'input_path': input_path,
         'output_path': output_path,
         'mode': mode.value,


ifsbench/config_mixin.py
```diff
@@ -39,67 +48,133 @@ def from_config(
...
+        candidates = [cls]
 
-        Args:
-            config: names and values for member variables.
+        while candidates:
+            current = candidates.pop(0)
 
-        Returns:
-            class instance
+            if AbstractDataClass in current.__bases__:
```
uhager, line 130: This quite neatly solves the problems I had with abstracting out the subclass code.
But will this be flexible enough if at some point we have another layer of inheritance (probably a bad idea anyway, so I guess we should avoid that issue if we can), e.g. maybe an SrunLauncher, and a DebugSrunLauncher, a ValgrindSrunLauncher or something like that? Could possibly get around issues like that with mixins or utils, but should at least make sure this won't be a blocker down the line.

johannesbulin: I guess that we could support multiple layers of inheritance here by adding a subclass not only to the `_subclasses`  dict of the "closest" abstract base class but also to other base classes higher up in the class hierarchy.

ifsbench/config_mixin.py
```diff
@@ -7,30 +7,39 @@
...
 # in implementing classes.
-CLASSNAME = 'classname'
+CLASSNAME = 'class_name'
 RESERVED_NAMES = [
     CLASSNAME,
 ]
 
 
-class ConfigMixin(ABC):
+class DataClass(BaseModel, use_enum_values=True):
```
uhager, line 32: I think I still prefer ConfigMixin to better describe what this is. DataClass sounds a bit too generic for me, especially since it's not just any data, it has to be something that can be converted to json/yaml. I don't really have any better ideas, will think about it some more.

ifsbench/config_mixin.py
```diff
@@ -39,67 +48,133 @@ def from_config(
...
+            'Allowed',
+            'Union[Dict[Allowed, Allowed], List[Allowed], str, int, float, bool, None]',
+        )
 
-class PydanticConfigMixin(ConfigMixin, BaseModel, use_enum_values=True):
+        allowed_type = TypeAdapter(Dict[str, Allowed])
+
+        return allowed_type.validate_python(config)
+
+class AbstractDataClass(DataClass):
```
uhager, line 91: Sorry to harp on the names so much, but I find this rather confusing: usually, you'd have an AbstractDataClass and then the implemented DataClass. Here, it's the other way round. Maybe call this SubclassableDataClass

uhager: So this is now SerialisationMixin, but it still has this unexpected inversion of which one is Abstract. Could you still rename the AbstractSerialisationMixin to something that indicates which part is abstract and that it is a type if SerialisationMixin, not the other way round?

ifsbench/results/ensemble_stats.py
```diff
@@ -5,92 +5,31 @@
...
-ENSEMBLE_DATA_KEY = 'ensemble_data'
 # Statistics keywords available when calling calc_stats. In addition
 # percentiles are supported with format '[p,P](\d{1,2})'.
 AVAILABLE_BASIC_STATS = ['min', 'max', 'mean', 'median', 'sum', 'std']
 
-_JSON_ORIENT = 'columns'
-
-
-class EnsembleStats(ConfigMixin):
+class EnsembleStats(SerialisationMixin):
```
uhager, line 24: It's been a while, but from what I remember this class had two different serialisation options: It can be created/serialised with a data file path, or it can serialise the actual data. If I'm reading this change correctly, the former option is no longer available, no more file path, only data frames? Is this intended? I'm just worried that at some point that data may get rather large and make the config file hard to read or we may want more efficient data formats than ascii.
If those aren't issues, and you think ascii as part of the config file will suffice, I'm ok with this, just want to make sure we're not losing some needed functionality.

johannesbulin: Yes, I pondered about this for a bit. Here are the reasons why I thought it might be better to remove it:

- Even if the path was given, the frames would still be written to a text file (I think your initial implementation dumped them into a JSON file). So file-size wise there is probably no big difference between dumping the ensemble to the given path and dumping the whole `EnsembleStats` object to YAML/JSON.
- It's obviously much easier to support the direct-dumping in pydantic. Still, if it becomes necessary, we can probably support the old path behaviour again (using custom model validators) - but dumping to a binary format, maybe.
- At least initially, we will only have small "stat"-like data frames, i.e. not the full data, only some statistics. _If_ this changes at some point, we can re-add this functionality again.

johannesbulin:
Rebased to current `main` branch to resolve conflicts.

PR #57, opened 2025-07-14
Removed IFS-specific input data command line tools.

johannesbulin:
Removed IFS specific command line tools in `ifsbench/command_line/pack.py`. Rational behind this:

- As this is application specific, this will move out of `ifsbench` and into the corresponding target repository (IFS specific one).
- Taking `pack.py` away, makes it possible to use the `cli` click-group without automatically including all the pack/unpack commands.


johannesbulin:
> Are the utilities already available somewhere? It would be good to keep functionality alive before removing it.

Not yet. My "plan" was as follows:

1. Finish work on IFS-specific components (in `ifs-test`). Let changes be reviewed.
2. Merge this PR (which removes the IFS-specific tools from ifsbench). Create a tag for the resulting code.
3. Use this tag in `ifs-test`.

reuterbal (approved):

PR #58, opened 2025-07-15
Added datahandler for namelist sanitisation.

johannesbulin:
Added an additional datahandler that can sanitise namelists (i.e. resolve duplicates). This is necessary for using this in IFS, as the default IFS namelists are usually not sanitised.

### DIFF
--- a/ifsbench/data/namelisthandler.py
+++ b/ifsbench/data/namelisthandler.py
@@ -166,6 +166,9 @@ class NamelistHandler(DataHandler):
         debug(f"Modify namelist {input_path}.")
         namelist = f90nml.read(input_path)
 
+        namelist.uppercase = True
+        namelist.end_comma = True
+
         for override in self.overrides:
             override.apply(namelist)
 
--- a/ifsbench/data/namelisthandler.py
+++ b/ifsbench/data/namelisthandler.py
@@ -17,7 +17,7 @@ import f90nml
 from ifsbench.config_mixin import PydanticConfigMixin
 from ifsbench.data.datahandler import DataHandler
 from ifsbench.logging import debug, info
-from ifsbench.namelist import sanitize_namelist
+from ifsbench.namelist import SanitiseMode, sanitise_namelist
 
 
 
@@ -145,7 +145,7 @@ class NamelistHandler(DataHandler):
     output_path: pathlib.Path
     overrides: List[NamelistOverride]
 
-    def execute(self, wdir, **kwargs):
+    def execute(self, wdir: pathlib.Path, **kwargs):
         wdir = pathlib.Path(wdir)
 
         if self.input_path.is_absolute():
@@ -180,27 +180,6 @@ class NamelistHandler(DataHandler):
 
         namelist.write(output_path, force=True)
 
-class SanitiseMode(Enum):
-    """
-    Specify, how duplicated entries in a namelist should be sanitised.
-    """
-
-    #: For multiply defined namelist groups, retain only the first.
-    FIRST = 'first'
-
-    #: For multiply defined namelist groups, retain only the last.
-    LAST = 'last'
-
-    #: For multiply defined namelist groups, merge variable definitions from
-    #: all groups. Conflicts are resolved by using the first occurence of a
-    #: variable.
-    MERGE_FIRST = 'merge_first'
-
-    #: For multiply defined namelist groups, merge variable definitions from
-    #: all groups. Conflicts are resolved by using the last occurence of a
-    #: variable.
-    MERGE_LAST = 'merge_last'
-
 class NamelistSanitiseHandler(DataHandler):
     """
     DataHandler specialisation that can sanitise Fortran namelists.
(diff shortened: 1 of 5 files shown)


ifsbench/data/namelisthandler.py
```diff
@@ -164,7 +166,86 @@ def execute(self, wdir, **kwargs):
...
+    output_path: pathlib.Path
+
+    #: The sanitising/merging strategy that is used. See :class:`SanitiseMode`
+    #: for the different options.
+    mode: SanitiseMode = SanitiseMode.MERGE_LAST
+
+    def execute(self, wdir, **kwargs):
+        wdir = pathlib.Path(wdir)
+
+        if self.input_path.is_absolute():
```
uhager, line 223: Not for this PR, but I think I've seen this absolute check in a few different classes, maybe we should pull it out into some utils, or could go into ConfigMixin or whatever it will be called. Again, not for this PR, but I can do that once your PRs are merged.

ifsbench/data/namelisthandler.py
```diff
@@ -164,7 +166,86 @@ def execute(self, wdir, **kwargs):
...
 
+        namelist.uppercase = True
+        namelist.end_comma = True
+
         for override in self.overrides:
             override.apply(namelist)
 
         namelist.write(output_path, force=True)
+
+class SanitiseMode(Enum):
```
uhager, line 177: If this is serialised, I think you need (str, Enum) to correctly print the strings in a nice format.

ifsbench/data/namelisthandler.py
```diff
@@ -164,7 +166,86 @@ def execute(self, wdir, **kwargs):
...
+    """
+    Specify, how duplicated entries in a namelist should be sanitised.
+    """
+
+    #: For multiply defined namelist groups, retain only the first.
+    FIRST = 'first'
```
uhager, line 183: Please update namelist.py to use the enums rather than the hard-coded strings. May need moving this Enum to avoid a circular dependency.

ifsbench/data/namelisthandler.py
```diff
@@ -164,7 +166,86 @@ def execute(self, wdir, **kwargs):
...
+    #: all groups. Conflicts are resolved by using the first occurence of a
+    #: variable.
+    MERGE_FIRST = 'merge_first'
+
+    #: For multiply defined namelist groups, merge variable definitions from
+    #: all groups. Conflicts are resolved by using the last occurence of a
+    #: variable.
+    MERGE_LAST = 'merge_last'
+
+class NamelistSanitiseHandler(DataHandler):
```
uhager, line 198: How does this compare to IFSNameList which also has sanitising? Can that be deleted?

ifsbench/data/namelisthandler.py
```diff
@@ -164,7 +166,86 @@ def execute(self, wdir, **kwargs):
...
+    #: The path to which the updated namelist will be written. If a relative
+    #: path is given, this will be relative to the ``wdir`` argument in
+    #: :meth:`execute`.
+    output_path: pathlib.Path
+
+    #: The sanitising/merging strategy that is used. See :class:`SanitiseMode`
+    #: for the different options.
+    mode: SanitiseMode = SanitiseMode.MERGE_LAST
+
+    def execute(self, wdir, **kwargs):
```
uhager, line 220: Please add type annotation for wdir

ifsbench/data/namelisthandler.py
```diff
@@ -164,7 +166,86 @@ def execute(self, wdir, **kwargs):
...
+            return
+
+        if self.output_path.is_absolute():
+            output_path = self.output_path
+        else:
+            output_path = wdir / self.output_path
+
+
+        debug(f"Sanitise namelist {input_path} using sanitise mode {self.mode}.")
+
```
uhager, line 240: Following up on my previous comment, there seems to be quite a bit of duplication with NamelistHandler. Could this be unified somehow? Make NamelistSanitiseHandler inherit from NamelistHandler and either split the duplicate code in the execute() function out into other functions that can the be assembled, or make a dedicated NamelistOverridesHandler, and the base NamelistHandler has the execute() function but then where it now applies the overrides or sanitises it has an abstract modify_namelist (or what you want to call it) function that is implemented in the subclasses.

johannesbulin: I would like to keep it how it is for the moment. If we get a third namelist-related class I'll try to think of a cleaner solution.

ifsbench/data/namelisthandler.py
uhager: You can assign an issue to me for cleanup, it's been a while since I've worked on ifsbench and I need to remind myself what's where.

johannesbulin: Added issue #59

reuterbal (approved):
I agree with @uhager 's comments. Otherwise this seems fine to me

uhager (approved):
Thanks for implementing the changes!

PR #63, opened 2025-08-20
Release/0.1.3

msleigh:


reuterbal:
I don't think we should merge 0.1.3 into main directly - main has already moved on substantially with the refactoring work of @johannesbulin and some of the changes that have been made for 0.1.3 are not compatible with the new way of specifying resources there.

But I'll take care of carrying forward the packaging and version locking

reuterbal:
Filed #65

PR #64, opened 2025-08-20
Add AtosAg arch

awnawab:
This PR contributes an arch for the new Atos ag partition.

### DIFF
--- a/ifsbench/arch.py
+++ b/ifsbench/arch.py
@@ -318,48 +318,17 @@ class Atos(Arch):
             if qos:
                 launch_user_options.insert(0, qos)
 
-            # To ensure we correctly bind GPUs to CPUs we write a little
-            # wrapper script that maps the correct CUDA_VISIBLE_DEVICES
-            # id to each rank locally per node.
-            # Because the mapping from NUMA domain to GPU is
-            # [1->0, 0->1, 3->2, 2->3], we need to apply some reordering here
+            wrapper_str = cls.get_gpu_binding_script(cpus_per_task, threads_per_core, tasks_per_node)
 
-            # The GPU that is closest to each core
-            core_to_gpu_mapping = {core: 1 for core in range(32)}
-            core_to_gpu_mapping.update({core: 0 for core in range(32, 64)})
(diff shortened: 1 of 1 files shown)


ifsbench/arch.py
```diff
@@ -416,6 +411,81 @@ class AtosAcCpuConfig(CpuConfiguration):
...
+        """Create script to bind GPUs to CPU cores."""
+
+        # To ensure we correctly bind GPUs to CPUs we write a little
+        # wrapper script that maps the correct CUDA_VISIBLE_DEVICES
+        # id to each rank locally per node.
+        # Because the mapping from NUMA domain to GPU is
+        # [1->0, 0->1, 3->2, 2->3], we need to apply some reordering here
+
+        # The GPU that is closest to each core
+        core_to_gpu_mapping = {core: 1 for core in range(32)}
```
johannesbulin, line 425: This part of the code is used by `AtosAc` as well as `AtosAg`. `AtosAg`, however, has a completely different CPU setup (more sockets, more cores per socket). This `core_to_gpu_mapping` seems to be tailored to `AtosAc` (128 cores per node) and will probably not work as intended with `AtosAg`, where we probably would use different ranges (multiples of 72, instead of 32).

johannesbulin (changes requested):
Most changes look good. I am not sure at the moment, that the GPU-binding-script will work on `AtosAg` as intended. Might be great if @reuterbal could have a look at this as well.

johannesbulin (approved):

ifsbench/arch.py
```diff
@@ -314,55 +314,28 @@ def run(cls, cmd, tasks, cpus_per_task, threads_per_core, launch_cmd=None,
                 gpus_per_node = min(gpus_per_task * tasks_per_node, cls.cpu_config.gpus_per_node)
```
awnawab, line 314: Why do we set the `tasks_per_node` on line 310? that currently prevents us on ag to run more than 4 tasks_per_node, since we always need to specify gpus on ag. If `gpus_per_node` is never allowed to exceed `cls.cpu_config.gpus_per_node` then it should be safe to not recompute `tasks_per_node` here.

PR #68, opened 2025-09-19
DrHook: Small fix to DrHookRecord parsing, triggered by odb tags

mlange05:
Apparently, ODB DrHook tags have the format `xxx_@1:codb.c`, for which we need to strip the trailing filename when parsing the thread number.


reuterbal (approved):
LGTM

Note to all: There are no DrHook tests at all at the moment

PR #71, opened 2025-11-03
Resolved pylint issues

johannesbulin:
- Removed deprecated entry in pylintrc, mirroring what was done in https://github.com/ecmwf-ifs/loki/commit/7f37d24f7ebf6e9763f25ce2dc8d386222860faa. Thanks @mlange05.
- Prevented pylint from complaining about an used-before-assignment issue in `gribfile.py` (code is fine, pylint simply doesn't understand the initialisation there).

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -33,13 +33,6 @@ load-plugins=
 # Pickle collected data for later comparisons.
 persistent=yes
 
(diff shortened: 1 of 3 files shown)


reuterbal (approved):
LGTM :shipit:

PR #76, opened 2025-12-09
MPI and Srun launchers now set OMP_NUM_THREADS automatically

johannesbulin:
Until now, the MPI/Srun launchers didn't set the `OMP_NUM_THREADS` environment variable as this is somewhat application specific (not all applications are bound to use OpenMP).

Still, after consultation with @mlange05 I guess that it's less error-prone if the launchers set `OMP_NUM_THREADS` automatically. If this leads to unwanted behaviour with some applications, these applications can just unset/update `OMP_NUM_THREADS`.

### DIFF
--- a/ifsbench/launch/mpirunlauncher.py
+++ b/ifsbench/launch/mpirunlauncher.py
@@ -96,6 +96,16 @@ class MpirunLauncher(Launcher):
                     )
                 )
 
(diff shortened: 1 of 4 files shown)


reuterbal (approved):
Looks good and I agree that it should be safe to do this when requesting a specific number of cpus per task

PR #77, opened 2025-12-11
Use --map-by in MpirunLauncher instead of --cpus-per-proc.

johannesbulin:
The `--cpus-per-proc` flag for `mpirun` has been deprecated for a while and support for this seems to have ended in some MPI implementations now. Therefore, I've tried to replace it by the modern `--map-by` approach.

@reuterbal : Could you double-check that I am not doing something stupid here? I am not an expert on doing MPI bindings - and I have to combine the multi-threading and the binding here.

@mlange05 : Would be great if you could verify that this works as intended.

### DIFF
--- a/ifsbench/launch/mpirunlauncher.py
+++ b/ifsbench/launch/mpirunlauncher.py
@@ -23,7 +23,6 @@ class MpirunLauncher(Launcher):
         'tasks': '-n {}',
         'tasks_per_node': '--npernode {}',
         'tasks_per_socket': '--npersocket {}',
-        'cpus_per_task': '--cpus-per-proc {}',
     }
 
     _bind_options_map = {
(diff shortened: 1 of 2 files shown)


ifsbench/launch/mpirunlauncher.py
```diff
@@ -51,10 +50,24 @@ def _get_distribution_options(self, job: Job) -> List[str]:
         ):
             warning('Specified remote distribution option ignored in MpirunLauncher')
 
-        if job.distribute_local is None or job.distribute_local in do_nothing:
+        # By default use core mapping. At least that's the default in the
+        # OpenMPI implementation
+        # (https://docs.open-mpi.org/en/main/man-openmpi/man1/mpirun.1.html#label-schizo-ompi-map-by).
+        map_by = 'core'
+
+        if job.distribute_local and job.distribute_local not in do_nothing:
```
reuterbal: I would add `None` to `do_nothing` above and then it should be sufficient to have
```suggestion
        if job.distribute_local not in do_nothing:
```

reuterbal (approved):
That looks fine to me. The map-by concept actually provides a pretty good way to have a more fine-grained mapping between ifsbench's conceptual options and launch command execution in the future.

Issue #86, opened 2026-02-09
validate_result_identical missing error check


johannesbulin:
### What happened?

In `ifsbench/validation/frame_close_validation.py` we do a check
```
        if not equal and mismatch:
            is_identical = False
```
in the end. This is not sufficient, as the case `if not equal and not mismatch` is not handled - which corresponds to two frames having different shape/indices. In this case, we should also set `is_identical = False` to cause the comparison to fail.

### What are the steps to reproduce the bug?

-

### Version

0.2.0

### Platform (OS and architecture)

-

### Relevant log output

```shell

```

johannesbulin:
Fixed in #88.

PR #88, opened 2026-02-10
#86: frame validation fixes

uhager:
fix validation to distinguish between different fram shape / indices and value mismatches. Add debug output of full frames.
Update tests to match.

### DIFF
--- a/ifsbench/validation/frame_close_validation.py
+++ b/ifsbench/validation/frame_close_validation.py
@@ -149,12 +149,7 @@ def validate_result_identical(
             error(
                 f"First mismatch at ({idx}, {col}): {frame.loc[idx,col]} != {frame_ref.loc[idx,col]}."
             )
-            for idx, col in mismatch:
-                debug(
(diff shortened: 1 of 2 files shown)


ifsbench/validation/frame_close_validation.py
```diff
@@ -134,15 +134,31 @@ def validate_result_identical(
...
+
+        if mismatch:
             is_identical = False
             idx, col = mismatch[0]
             error(
                 f"First mismatch at ({idx}, {col}): {frame.loc[idx,col]} != {frame_ref.loc[idx,col]}."
             )
-
             for idx, col in mismatch:
                 debug(
```
johannesbulin: I think it would make sense to delete this part and just have the pure pandas-based output below.

mlange05: I would agree here. A `pd.DataFrame` has a pretty good textual representation and simply printing the relative error as `frame_ref - frame` tends to yield a lot of practical information for larger frames. In addition, the max relative error might also be a useful metric to print?

PR #92, opened 2026-02-20
Fixed bugs in BashLauncher

johannesbulin:
Stumbled over a couple of bugs in the `BashLauncher` when testing it in the real world... Fixed these and improved the testing.

- The bash script didn't set `cd rundir` which caused issues when running the script manually.
- Fixed handling/escaping of certain characters inside environment variables (escape ', $, ", ...).
- The `BashLauncher` will now ignore pseudo-variables with non-bash compatible names that still pop up in `os.environ`. For example, in my environment I see bash-defined functions like `BASH_FUNC_module%%` which are not environment variables and which can't be exported.

### DIFF
--- a/ifsbench/launch/bashlauncher.py
+++ b/ifsbench/launch/bashlauncher.py
@@ -60,8 +60,11 @@ class BashLauncher(LauncherWrapper):
                 debug(f'Ignore non-standard bash name {key} in BashLauncher.')
                 continue
 
+            if value is None:
+                continue
+
             # We set the environment variables inside a "..." block. This means
-            # that we have to escpae certain special characters (", $, `).
+            # that we have to escape certain special characters (", $, `).
             escaped_value = value.replace('$', '\\$')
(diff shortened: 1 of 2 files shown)


ifsbench/launch/bashlauncher.py
```diff
@@ -23,12 +25,54 @@ class BashLauncher(LauncherWrapper):
...
+            # Check that the current "environment variable" is an actual
+            # variable. For example, bash function definitions may be in
+            # `launch_data.env` but end with `_%%` and are not proper
+            # environment variables. We skip these.
+            if not env_name_check.fullmatch(key):
+                debug(f'Ignore non-standard bash name {key} in BashLauncher.')
+                continue
+
+            # We set the environment variables inside a "..." block. This means
+            # that we have to escpae certain special characters (", $, `).
```
uhager: nit: type: escape

uhager (approved):

PR #94, opened 2026-02-23
Add populate script

awnawab:
This PR contributes the populate script that allows the build-time python dependencies to be pre-fetched.

### DIFF
--- a/.github/workflows/pytest.yaml
+++ b/.github/workflows/pytest.yaml
@@ -77,12 +77,17 @@ jobs:
       matrix:
         platform: ['ubuntu-latest']
         python-version: ['3.8', '3.9', '3.10', '3.11', '3.12', '3.13']
+        populate: [true, false]
 
         include:
           - platform: 'macos-latest'
             python-version: '3.11'
+            populate: true
+          - platform: 'macos-latest'
+            python-version: '3.11'
+            populate: false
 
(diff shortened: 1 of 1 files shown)


cmake/ifsbench_download_python_wheels.cmake
```diff
@@ -0,0 +1,142 @@
...
+    if( "${_RET_VAL}" EQUAL "0" )
+
+        message( STATUS "All dependency wheels for ${_PAR_REQUIREMENT_SPEC} found in cache" )
+
+    else()
+
+        message( STATUS "Downloading dependency wheels for ${_PAR_REQUIREMENT_SPEC} to ${WHEELS_DIR}" )
+
+        # Download build dependencies separately, since these aren't automatically recongnized by pip
+        execute_process(
```
johannesbulin: Would it make sense to raise an error/quit when this operation fails? At the moment there is no check that this actually works.

johannesbulin (changes requested):
Looks very nice in general! As this will probably be used in multiple places in production, I would like to request that the `populate` feature gets tested automatically to avoid any kind of nasty surprises.

Probably easiest to run all of the `pytest` tests in "standard mode" and "populate first, build afterwards" mode.

cmake/ifsbench_download_python_wheels.cmake
awnawab: Great catch! will do

PR #102, opened 2026-03-04
Never use `--map-by core` when multiple threads are used.

johannesbulin:
I ran into a problem with the `MpirunLauncher` when used with `OpenMPI`. If no binding was specified (or the default was used) **and** `cpus_per_task>1`, the resulting launch command was something like
```
mpirun -n 2 --map-by core:PE=4 executable
```
This works with Intel MPI but not with OpenMPI. Some digging (and I am not 100% sure about this) showed that it is apparently not okay to map by core when using multiple threads. Some information here: https://github.com/open-mpi/ompi/issues/7717.

To get this to work, I replace all `--map by-core:PE` calls by `--map-by slot:PE` if more than one thread is requested.


uhager (approved):

reuterbal:
This looks good to me. Does this handle the oversubscribe situation correctly as well?

reuterbal (approved):

PR #108, opened 2026-03-18
New PerturbationHandler to add random noise to netcdf files.

uhager:
Two types of noise are available, Gaussian and uniform; others can be added.
The user has to specify the size of the noise since there isn't really a good way to know what is reasonable based on the netcdf metadata, at least nothing general.
The user has the option to apply the same noise to all variables, or pass in a dictionary of variables and perturbations to apply.
A random number seed can be passed in to make testing reproducible, otherwise the numpy default is used.

The changes to netcdf_file_reader and data_file_stats are needed since calling to_netcdf by the PerturbationHandler causes the tests to fail. It seems that the exception from netcdf / hdf5 varies depending on whether hdf5 was already used, which is triggered by calling xarray.dataset.to_netcdf. This makes the outcome of the tests order dependent, if PerturbationHandler is tested first, the other tests fail, if not they pass. Adding our own message makes the error clearer and consistent.


johannesbulin (changes requested):

johannesbulin (changes requested):
Just one minor typo - otherwise it looks good! Thanks for the work!

johannesbulin (approved):

PR #112, opened 2026-05-19
YAML extension

johannesbulin:
### Description

- Added central support in `ifsbench` to read YAML files.
- Added support for combining multiple YAML files by adding support for an `!import other_file` YAML constructor.
- Added support for simple parameterisation like this:
```
templates:
  greeting:
    message: 'Hello, ${name}!'
 
instances:
  welcome: !configure:templates/greeting
    name: World
```

### Contributor Declaration

### DIFF
--- a/ifsbench/tests/test_yaml.py
+++ b/ifsbench/tests/test_yaml.py
@@ -22,8 +22,6 @@ def yaml_dir(tmp_path):
     return tmp_path
 
 
-# -- basic loading -----------------------------------------------------------
-
 def test_read_yaml_basic(yaml_dir):
     """read_yaml loads a simple YAML file."""
     f = yaml_dir / 'basic.yaml'
@@ -39,6 +37,7 @@ def test_read_yaml_file_not_found():
 
 
 # -- !import -----------------------------------------------------------------
+
 
 def test_import_basic(yaml_dir):
     """!import includes another YAML file."""
@@ -76,12 +75,13 @@ def test_import_nested(yaml_dir):
 
 # -- !configure --------------------------------------------------------------
 
+
 def test_configure_basic(yaml_dir):
     """!configure substitutes template placeholders."""
     content = textwrap.dedent("""\
         templates:
           greeting:
-            message: Hello, {name}!
+            message: Hello, ${name}!
 
         instances:
           welcome: !configure:templates/greeting
@@ -97,8 +97,8 @@ def test_configure_multiple_values(yaml_dir):
     content = textwrap.dedent("""\
         templates:
           point:
-            x: "{x}"
-            y: "{y}"
+            x: ${x}
+            y: "${y}"
 
         instances:
           origin: !configure:templates/point
@@ -107,7 +107,7 @@ def test_configure_multiple_values(yaml_dir):
     """)
     (yaml_dir / 'cfg.yaml').write_text(content)
     result = read_yaml(str(yaml_dir / 'cfg.yaml'))
-    assert result['instances']['origin'] == {'x': '0', 'y': '0'}
+    assert result['instances']['origin'] == {'x': 0, 'y': 0}
 
 
 def test_configure_preserves_template(yaml_dir):
@@ -115,7 +115,7 @@ def test_configure_preserves_template(yaml_dir):
     content = textwrap.dedent("""\
         templates:
           item:
-            value: "{value}"
+            value: ${value}
 
         instances:
           a: !configure:templates/item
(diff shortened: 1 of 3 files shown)


ifsbench/yaml.py
```diff
@@ -0,0 +1,177 @@
...
+        return {k: _resolve_markers(v, root) for k, v in obj.items()}
+    if isinstance(obj, list):
+        return [_resolve_markers(item, root) for item in obj]
+    return obj
+
+
+def read_yaml(filename: Union[str, Path], encoding='utf-8') -> Dict[str, Any]:
+    """
+    Parse a YAML file and return the resulting dictionary.
+
```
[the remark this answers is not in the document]
johannesbulin: Switched to `Any`.

ifsbench/yaml.py
```diff
@@ -0,0 +1,177 @@
...
+        """Handle ``!import other_file.yaml``."""
+
+        # Load the import path.
+        rel_path = loader.construct_scalar(node)
+
+        # Only use the path relative to the base dir.
+        import_path = (base_dir / rel_path).resolve()
+
+        if not import_path.is_file():
+            raise FileNotFoundError(f'Imported YAML file not found: {import_path}')
```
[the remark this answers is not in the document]
johannesbulin: Fixed.

ifsbench/yaml.py
```diff
@@ -0,0 +1,199 @@
...
+    -------
+        included: !import other_file.yaml
+
+        templates:
+          greeting:
+            message: Hello, ${name}!
+
+        instances:
+          welcome: !configure:templates/greeting
+            name: World
```
reuterbal, line 176: Note that usually nothing after parameters/returns is rendered - so I  suspect you have to move this further up in the docstring.

reuterbal (commented):
This looks very sophisticated and thought-through. I cannot claim I fully understood every detail but Pyyaml's add_constructor functionality is the right way of installing these custom handlers (we've been using a different implementation via the same hook in Loki for a while: https://github.com/ecmwf-ifs/loki/blob/main/loki/tools/util.py#L454).

Just one question regarding the keyword choice: We've been using `!include` and it's what I've encountered in several other cases as keyword. Is there a specific reason for using `!import` here or would it be sensible to follow what seems to me (happy to be proven wrong, of course) like the more widespread choice?

ifsbench/yaml.py
johannesbulin: Thanks, will have a look!

johannesbulin:
> Just one question regarding the keyword choice: We've been using `!include` and it's what I've encountered in several other cases as keyword. Is there a specific reason for using `!import` here or would it be sensible to follow what seems to me (happy to be proven wrong, of course) like the more widespread choice?

I am happy to switch to `!include` instead, to align with what we do elsewhere. I guess I've been using too much Python in recent years so I am trying to `import` everything now...

ifsbench/tests/test_yaml.py
```diff
@@ -0,0 +1,238 @@
...
+def test_configure_multiple_values(yaml_dir):
+    """!configure substitutes multiple placeholders."""
+    content = textwrap.dedent("""\
+        templates:
+          point:
+            x: ${x}
+            y: "${y}"
+
+        instances:
+          origin: !configure:templates/point
```
mlange05, line 134: The test here doesn't actually test the template engine, as it replaces values like-for-like with identical keys. As a result, you can happily remove the `!configure ...` code here and the test will still pass. I think this needs some tightening to actually test that the template-resolution works as expected.

The same applies for two more tests below.

ifsbench/tests/test_yaml.py
```diff
@@ -0,0 +1,238 @@
...
+
+def test_configure_preserves_template(yaml_dir):
+    """!configure does not mutate the original template."""
+    content = textwrap.dedent("""\
+        templates:
+          item:
+            value: ${value}
+
+        instances:
+          a: !configure:templates/item
```
mlange05, line 152: See above comment

ifsbench/tests/test_yaml.py
```diff
@@ -0,0 +1,238 @@
...
+
+def test_configure_nested_template(yaml_dir):
+    """!configure works with deeply nested template paths."""
+    content = textwrap.dedent("""\
+        some_template:
+          default_template:
+            value: ${value}
+
+        instances:
+          my_instance: !configure:some_template/default_template
```
mlange05, line 185: See above comment.

mlange05 (commented):
A very nice extension, and the implementation looks good to me. 

I've flagged some slightly lax testing that could probably be improved (the functionality was immediately clear from reading the tests), but otherwise this looks good to me.

ifsbench/yaml.py
johannesbulin: Should be fixed now.

ifsbench/tests/test_yaml.py
johannesbulin: Yes, good point. Just fixed these tests.

mlange05 (approved):
Thanks for addressing the comment. Looks great now! GTG 👍

PR #115, opened 2026-05-30
Modernize ifsbench CMake Python installation (fixes #114)

### LINKED ISSUE #114: IFSbench is not CMake-installable
### What happened?

IFSbench can be downloaded and installed into a virtual environment as a build-action via CMake. However, it cannot be installed to an installation path.

### What are the steps to reproduce the bug?

```
git clone https://github.com/ecmwf-ifs/ifsbench
cmake -S ifsbench -B build
cmake --build build
cmake --install build --prefix install_dir
```
This creates a virtual environment in the `build` directory but does not actually make ifsbench available in the `install_dir`:
```
$ tree install_dir/
install_dir/
├── include
└── lib64
    └── cmake
        └── ifsbench
            ├── ifsbench-config.cmake
            └── ifsbench-config-version.cmake
```

### Version

main

### Platform (OS and architecture)

Linux ad6-198.bullx 4.18.0-553.105.1.el8_10.x86_64 #1 SMP Sat Feb 7 00:49:44 EST 2026 x86_64 x86_64 x86_64 GNU/Linux

### Relevant log output

```shell

```

### Accompanying data

_No response_

### Organisation

_No response_

reuterbal:
### Description

- Rework ifsbench CMake installation to use Loki-style Python venv/package macros, configure-time package installation, install-time prefix venv population, and exported CLI targets.
- Add ifsbench package export/import support so downstream `ecbuild_find_package(ifsbench)` users can access `IFSBENCH_Python3_EXECUTABLE` and imported CLI targets from both build-tree and installed packages.
- Add Loki-style wheel population support via `populate` and `ifsbench_get_python_wheels.cmake`, including offline install validation.
- Add pytest coverage for CMake configure/install, prefix availability, downstream package discovery, and populated wheelhouse installation.

NB: This targets release 0.1.4 for inclusion in 49r3. I will rebase and file the same against main once this PR has been merged.

### Contributor Declaration


populate
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+
+# (C) Copyright 2024- ECMWF.
```
johannesbulin: Every other file has Copyright 2020- here.
```suggestion
# (C) Copyright 2020- ECMWF.
```

tests/test_cmake.py
```diff
@@ -0,0 +1,180 @@
...
+def fixture_ecbuild_cmake_dir(tmp_path_factory):
+    """
+    Download ecbuild and return its CMake module directory.
+    """
+    if not shutil.which('git'):
+        pytest.skip('git not available')
+
+    tmp_dir = tmp_path_factory.mktemp('ecbuild')
+    ecbuild_dir = tmp_dir/'ecbuild'
+    execute(['git', 'clone', 'https://github.com/ecmwf/ecbuild.git', str(ecbuild_dir)])
```
johannesbulin: This will always give us the `main` branch of `ecbuild` which may screw things up. I'd guess it might be better to use a fixed ecBuild tag here.

tests/test_cmake.py
```diff
@@ -0,0 +1,180 @@
...
+    cmake_args = []
+
+    if request.param:
+        artifacts_dir = tmp_dir/'artifacts'
+        env['ARTIFACTS_DIR'] = str(artifacts_dir)
+        execute([str(repo_root/'populate')], cwd=repo_root, env=env)
+        assert any(artifacts_dir.glob('*.whl'))
+
+        cmake_args += [f'-DARTIFACTS_DIR={artifacts_dir}']
+        env['http_proxy'] = 'http://foo.bar.baz'
```
johannesbulin, line 83: Why do we set the proxy here? And to foo bar?

johannesbulin (changes requested):
Thanks for all the work! I've tested the new `ifsbench` build process standalone and it seems to work nicely. When I use this `ifsbench` version in an IFS bundle build, however, something in the Python installation seems weird:

- During the bundle build, the `build_dir/ifsbench/ifsbench_env` venv is created. Fine so far.
- When the actual `ifsbench` Python module is installed, it doesn't end up in this venv but in `~/.local/lib/pythonXX/site-packages`.
- The same thing happens when doing the `install` step - the installed venv in `install/var/ifsbench_venv` doesn't contain the `ifsbench` Python module as this goes into `~/.local`.

Is this a bug in the `ifsbench` CMake or just a not-updated-IFS-bundle?

tests/test_cmake.py
reuterbal: Good point, yes. I'd suggest using the minimum required ecbuild version, which helps us spot if that lower boundary is no longer sufficient. Incorporating also the copilot suggestion:

```suggestion
    ecbuild_version = '3.7.0'
    execute(['git', 'clone', '--depth', '1', '-b', ecbuild_version, 'https://github.com/ecmwf/ecbuild.git', str(ecbuild_dir)])
```

tests/test_cmake.py
reuterbal: The point of the separate populate step is to allow the pip install to work even if we are on a system without internet access. To verify that this is effective, I need to prevent the use of packages from pypi - and because pip honours the http{s}_proxy variables we are faking no internet access by routing all requests via an url that cannot work

reuterbal:
Many thanks for the diligent review @johannesbulin . There was indeed some form of cache pollution breaking the venv installation mechanics, with cache variables preventing the macros from picking up the venv python interpreter. This only occurs in superbuilds (like ecbundle builds) where another project had looked for python already - and we have many more of these accrue to our projects recently. In my testing this appears to be fixed now, so this should be good for another look.

I have also fixed the various other comments and suggestions, all good ones.

johannesbulin (changes requested):
Thanks for all the updates. Most of the stuff seems to be working now. I've just encountered one issue when using this version of `ifsbench` in `ifs-source`. "Normal" builds work well but if I run `ifs-bundle populate` before the build, I get this weird error:
```
​- Installing Python package .../ifs-bundle/source/ifsbench
'.../ifs-bundle/build/ifsbench/ifsbench_env/bin/python3' '-m' 'pip' 'i
nstall' '--no-index' '--find-links=.../ifs-bundle/source/artifacts' '.../ifs-bundle/source/ifsbench'
Looking in links: .../ifs-bundle/source/artifacts
Processing .../ifs-bundle/source/ifsbench
  Installing build dependencies: started
  Installing build dependencies: finished with status 'done'
  Getting requirements to build wheel: started
  Getting requirements to build wheel: finished with status 'done'
  Preparing metadata (pyproject.toml): started
  Preparing metadata (pyproject.toml): finished with status 'done'
Processing .../ifs-bundle/source/artifacts/f90nml-1.5.0-py2.py3-none-any.whl (from ifsbench==0.1.dev1+g10f6abb79.d20260603)
INFO: pip is looking at multiple versions of ifsbench to determine which version is compatible with other requirements. This could take a while.
ERROR: Could not find a version that satisfies the requirement pandas (from ifsbench) (from versions: none)
ERROR: No matching distribution found for pandas
CMake Error at ifsbench/cmake/ifsbench_python_macros.cmake:422 (execute_process):
  execute_process failed command indexes:

    1: "Child return code: 1"

Call Stack (most recent call first):
  ifsbench/CMakeLists.txt:116 (ifsbench_install_python_package)
```
Any idea where this comes from?

johannesbulin:
> Thanks for all the updates. Most of the stuff seems to be working now. I've just encountered one issue when using this version of `ifsbench` in `ifs-source`. "Normal" builds work well but if I run `ifs-bundle populate` before the build, I get this weird error:
> 
> ```
> ​- Installing Python package .../ifs-bundle/source/ifsbench
> '.../ifs-bundle/build/ifsbench/ifsbench_env/bin/python3' '-m' 'pip' 'i
> nstall' '--no-index' '--find-links=.../ifs-bundle/source/artifacts' '.../ifs-bundle/source/ifsbench'
> Looking in links: .../ifs-bundle/source/artifacts
> Processing .../ifs-bundle/source/ifsbench
>   Installing build dependencies: started
>   Installing build dependencies: finished with status 'done'
>   Getting requirements to build wheel: started
>   Getting requirements to build wheel: finished with status 'done'
>   Preparing metadata (pyproject.toml): started
>   Preparing metadata (pyproject.toml): finished with status 'done'
> Processing .../ifs-bundle/source/artifacts/f90nml-1.5.0-py2.py3-none-any.whl (from ifsbench==0.1.dev1+g10f6abb79.d20260603)
> INFO: pip is looking at multiple versions of ifsbench to determine which version is compatible with other requirements. This could take a while.
> ERROR: Could not find a version that satisfies the requirement pandas (from ifsbench) (from versions: none)
> ERROR: No matching distribution found for pandas
> CMake Error at ifsbench/cmake/ifsbench_python_macros.cmake:422 (execute_process):
>   execute_process failed command indexes:
> 
>     1: "Child return code: 1"
> 
> Call Stack (most recent call first):
>   ifsbench/CMakeLists.txt:116 (ifsbench_install_python_package)
> ```
> 
> Any idea where this comes from?

Was basically my own fault by running populate with a different Python version than the build. Using the same version fixes it.

cmake/ifsbench_get_python_wheels.cmake
```diff
@@ -0,0 +1,23 @@
+# (C) Copyright 2024- ECMWF.
```
johannesbulin:
```suggestion
# (C) Copyright 2020- ECMWF.
```
One last license header change, then I am happy.

cmake/ifsbench_get_python_wheels.cmake
```diff
@@ -0,0 +1,23 @@
...
+# In applying this licence, ECMWF does not waive the privileges and immunities
+# granted to it by virtue of its status as an intergovernmental organisation
+# nor does it submit to any jurisdiction.
+cmake_minimum_required( VERSION 3.19 FATAL_ERROR )
+
+list( APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_LIST_DIR}" )
+include( ifsbench_python_macros )
+
+set( WHEELS_DIR "${CMAKE_CURRENT_BINARY_DIR}/wheels" CACHE PATH "" )
+set( WHEELS_DIR "${CMAKE_CURRENT_BINARY_DIR}/wheels" CACHE PATH "" )
```
johannesbulin, line 13: This line is duplicated. Can you remove it?

johannesbulin (commented):
Can confirm that it works now! Two minor style issues are left - afterwards I am happy to merge it.

cmake/ifsbench_get_python_wheels.cmake
reuterbal: Good spot, thanks!
```suggestion
```

johannesbulin (approved):

PR #117, opened 2026-06-16
CMake ifsbench installation

johannesbulin:
### Description

Cherry-picked changes from #115 into main. The main motivation was to have the build-venv-at-configure-time feature as this allows us to read YAML files with `!include` or `!configure` statements during configuration time.

### Contributor Declaration


ifsbench/gribfile.py
```diff
@@ -29,7 +29,7 @@
         raise ImportError('eccodes version too low.')
     # pylint: enable=no-member
     ECCODES_AVAILABLE = True
-except (RuntimeError, ImportError):
+except (RuntimeError, ImportError, AttributeError):
```
johannesbulin, line 32: Had to add this here to support ifs-bundle builds. If `ifsbench` is used at configure time (which is possible now), an `eccodes` Python module is available but not fully set up as this happens at build time. Therefore `eccodes` is defined but `eccodes.__version__` is not.

.gitignore
```diff
@@ -9,3 +9,5 @@ ifsbench.egg-info/*
 
 # Auto-generated docs files
 /docs/source/ifsbench.*
+
+artifacts/
```
reuterbal: Preferably only exclude the artifacts directory in the repository root to prevent confusion if something was to be called "artifacts" later own
```suggestion
/artifacts
```

cmake/ifsbench_get_python_wheels.cmake
reuterbal, line 1: The already existing `cmake/ifsbench_download_python_wheels.cmake` file should then be removed.

reuterbal (changes requested):
Sorry for dragging my feet on this and many thanks for taking this over. I think this looks good - just two small remarks.

cmake/ifsbench_get_python_wheels.cmake
johannesbulin: Well spotted, thanks!

reuterbal (approved):

PR #118, opened 2026-06-17
Updated version to 0.2.3

johannesbulin:
New tag for the latest updates which include

- Improved YAML support (#112) - a new `read_yaml` function is available that supports the tags `!include` (include data from another YAML file) and`!configure` (parameterise a given template).
- Updated CMake/ecBuild installation process (#117). An ifsbench Python virtual environment will now be created and populated at configure time, not at build time.
- Added a `PerturbationHandler` (#108) that can add noise to NetCDF files.


mlange05 (approved):
Looks great, many thanks! 👍 :shipit:

